PR #1, opened 2019-07-01
Make using old PIP easier

nedrebo:
This change improves the diagnostics from setup.py. A new helpful
error messages will trigger if any dependencies are missing when
installing the package. This typically happens when PIP version is <19
and the user is missing the build dependencies.

Example error message:

    Missing module 'skbuild'. Please install 'scikit-build' manually
    or use PIP>=19 to handle build dependencies automatically (PEP
    517).

Version number is bumped to 0.9.1 and this version will be uploaded to
PyPI.

The README.md is now simplified to not use venv since we gracefully
handle older PIP version. It is also updated to install the package
from PyPI and not git.

The Versioning section of README.md is extended to explain what all
those digits in the module version is all about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,8 @@
 
 Zivid Python is the official Python package for Zivid 3D cameras. Read more about Zivid at [zivid.com](https://www.zivid.com/).
 
-[![Build Status](https://img.shields.io/azure-devops/build/zivid-devops/376f5fda-ba80-4d6c-aaaa-cbcd5e0ad6c0/2/master.svg)](https://dev.azure.com/zivid-devops/zivid-python/_build/latest?definitionId=2&branchName=master)
+[![Build Status](https://img.shields.io/azure-devops/build/zivid-devops/376f5fda-ba80-4d6c-aaaa-cbcd5e0ad6c0/2/master.svg)](https://dev.azure.com/zivid-devops/zivid-python/_build/latest?definitionId=2&branchName=master) [![PyPI Package](https://img.shields.io/pypi/v/zivid.svg)](https://pypi.org/project/zivid/)
+
 
 <figure><p align="center"><img src="https://www.zivid.com/hs-fs/hubfs/images/www/ZividOnePlus.jpg?width=500&name=ZividOnePlus.jpg"></p></figure>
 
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,4 +1,6 @@
 [build-system]
+# This list is duplicated in setup.py
+# Keep the two lists in sync
```
trym-b: Can we use CMake to configure this list for both toml and setup.py?

setup.py
```diff
@@ -29,22 +26,50 @@ def _zivid_python_version():
...
-    install_requires=["numpy"],
-    cmake_args=[
-        "-DZIVID_PYTHON_VERSION=" + _zivid_python_version(),
-        "-DZIVID_SDK_VERSION=" + _zivid_sdk_version(),
-        "-Dpybind11_DIR=src/3rd-party/pybind11-2.2.4/share/cmake/pybind11/",
-    ],
-)
+def _check_dependency(module_name, package_hint=None):
+    try:
+        __import__(module_name)
```
trym-b:
```
from pkgutil import iter_modules
if "numpy" in [p.name for p in iter_modules()]:
    print("found")
else:
    print("not found")
```

trym-b: Then we don't accidentally import modules that are not needed at the moment.

setup.py
```diff
@@ -29,22 +26,50 @@ def _zivid_python_version():
...
+            )
+        )
+
+
+def _main():
+    # This list is a duplicate of the build-system requirments in pyproject.toml.
+    # The purpose of these checks is to help users with PIP<19 lacking support for
+    # pyproject.toml
+    # Keep the two lists in sync
+    _check_dependency("skbuild", "scikit-build")
```
trym-b: Can we configure these?

README.md
```diff
@@ -75,7 +71,39 @@ Standalone example programs can be found in the [samples](samples) directory.
...
 ## Versioning
 
-This python module is using [PEP 440](https://www.python.org/dev/peps/pep-0440) for versioning. The features available in the module depends on the Zivid SDK version used  when building the module. When updating this Python package it is *reccomended* to also update to the latest [Zivid SDK](http://www.zivid.com/software). Refer to the [Test Matrix](#test-matrix) for supported version.
+This python module is using [PEP 440](https://www.python.org/dev/peps/pep-0440) for versioning. The features available in the module depends on the Zivid SDK version used when building the module. When updating this Python package it is *recommended* to also update to the latest [Zivid SDK](http://www.zivid.com/software). Refer to the [Test Matrix](#test-matrix) for supported version.
+
+The version number of the Zivid Python module consists of six digits. The three first digits of the version is the [semantic version](https://semver.org/) of the code in this repository. The last three digits is the version of the underlying Zivid SDK library used by the Python module.
+
+### Version breakdown
+
+                                        Zivid SKD version = 1.4.1 (semantic version)
```
oysteinskotheim: Zivid SKD => Zivid SDK

README.md
```diff
@@ -23,29 +24,24 @@ Zivid Python is the official Python package for Zivid 3D cameras. Read more abou
 
 *Windows users also needs to make sure that the Zivid SDK installation folder is in system `PATH` before using the package, not only the terminal PATH variable. The default install location that should be added to system `PATH` is `C:\Program Files\Zivid\bin`.*
```
oysteinskotheim: Is it a possibility to copy the required DLLs from C:\Program Files\Zivid\bin to the location where the pyd file is located to avoid having to modify PATH?

nedrebo: Not that I am aware of. When and how do you propose to do it?

nedrebo:
> Review done, looks good. A few comments about the language, also a bit confused about the naming of the package: zivid-python or Zivid-Python or Zivid Python or zivid python

# Terminology used
Zivid Python: _the name of the package_
zivid-python: _the name of the repository_
Zivid-Python: _not used_
zivid: _the name of the main module in Zivid Python and the name we use when we upload Zivid Python to PyPI. Therefor we refer to it as "zivid module" or "zivid package"_

If you have a concrete proposal for better terminology please share it.

setup.py
nedrebo: No chicken egg problem, see PEP 517.

PR #3, opened 2019-07-02
Bump supported Zivid SDK version to 1.4.0

nedrebo:


setup.py
```diff
@@ -57,7 +57,7 @@ def _main():
         url="https://www.zivid.com",
         author="Zivid AS",
         author_email="support@zivid.com",
-        license=open("LICENSE").read(),
+        license="BSD 3-Clause",
```
trym-b: So in the future, if we want to re-license our code we then need to change the code as well?

nedrebo: Yes :(

If we think that's a problem we can use https://github.com/licensee/licensee. This is what github uses to detect license names. I do not think it is worth the effort.

PR #6, opened 2019-07-04
Raise a customized nested exception when _zivid fails at import

nedrebo:
The stacktrace from `import zivid` will now look like this:

```
Traceback (most recent call last):
  File "/tmp/env/lib/python3.7/site-packages/_zivid/__init__.py", line 3, in <module>
    from _zivid._zivid import (  # pylint: disable=import-error,no-name-in-module
ImportError: libZividCore.so: cannot open shared object file: No such file or directory

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/tmp/env/lib/python3.7/site-packages/zivid/__init__.py", line 8, in <module>
    from zivid.application import (
  File "/tmp/env/lib/python3.7/site-packages/zivid/application.py", line 2, in <module>
    import _zivid
  File "/tmp/env/lib/python3.7/site-packages/_zivid/__init__.py", line 23, in <module>
    ) from ex
ImportError: Failed to import the Zivid Python C-module, please check that:
 - Zivid SDK is installed
 - Zivid SDK version is matching the Zivid Python version
 - Zivid SDK libraries location is in PATH
```

### DIFF
--- a/modules/_zivid/__init__.py
+++ b/modules/_zivid/__init__.py
@@ -1,15 +1,32 @@
 """This file imports used classes, modules and packages."""
-from _zivid._zivid import (  # pylint: disable=import-error,no-name-in-module
-    __version__,
-    Application,
-    Camera,
-    CameraState,
-    environment,
-    firmware,
-    Frame,
-    FrameInfo,
-    hdr,
-    PointCloud,
-    Settings,
-    version,
-)
+try:
+    from _zivid._zivid import (  # pylint: disable=import-error,no-name-in-module
+        __version__,
+        Application,
+        Camera,
+        CameraState,
+        environment,
+        firmware,
+        Frame,
+        FrameInfo,
+        hdr,
+        PointCloud,
+        Settings,
+        version,
+    )
+except ImportError as ex:
+    import platform
+
+    def __missing_sdk_error_message():
+        error_message = """Failed to import the Zivid Python C-module, please verify that:
+ - Zivid SDK is installed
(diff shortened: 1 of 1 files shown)


nedrebo:
The original situation does not look to bad in my example, but on Windows it didn't spell out the name of libZividCore. There it only print that dll failed to load, making the error message very user un-friendly.

modules/_zivid/__init__.py
```diff
@@ -1,15 +1,23 @@
...
+        Frame,
+        FrameInfo,
+        hdr,
+        PointCloud,
+        Settings,
+        version,
+    )
+except ImportError as ex:
+    raise ImportError(
+        """Failed to import the Zivid Python C-module, please check that:
```
trym-b: check -> verify?

modules/_zivid/__init__.py
```diff
@@ -1,15 +1,23 @@
...
+ - Zivid SDK is installed
+ - Zivid SDK version is matching the Zivid Python version
```
trym-b: Zivid SDK is matching the Zivid Python version? So 0.9.0 should match 1.4.0?

modules/_zivid/__init__.py
```diff
@@ -1,15 +1,23 @@
...
+ - Zivid SDK libraries location is in PATH"""
```
trym-b: System PATH?

PR #16, opened 2019-09-08
Bump supported Zivid SDK version to 1.5.0

nedrebo:


knatten:
I notice you put this up as a draft PR, but still requested review, I take it we can start reviewing it then?

nedrebo:
> I notice you put this up as a draft PR, but still requested review, I take it we can start reviewing it then?

I am not sure, but I would say no. Draft means the PR will change and possibly invalidate all your comments and wasting the time you spent reviewing it.

The button that converts the draft into final version is called "[Ready for review]". So, I think that the intended workflow is for me to use the draft as a preview to make sure the PR is perfect before hitting that button.

PR #18, opened 2019-10-01
Make Settings.__str__() output a valid YAML string

eskaur:
With some small changes, the string-representation of a Settings
object can be made to be both valid YAML and more human-readable.

The string-representation of a Settings can now e.g. be dumped to
a human-readable file and subsequently be parsed into a nested
dictionary with a simple yaml.load().

### Sample:
```
import zivid
s = zivid.Settings()
print(s)
```
#### Old result:
```
Settings:
bidirectional: False
blue_balance: 1.081
brightness: 1.0
exposure_time: 0:00:00.008333
filters: Filters:
contrast: Contrast:
enabled: True
threshold: 5.0
outlier: Outlier:
enabled: True
threshold: 5.0
saturated: Saturated:
enabled: True
reflection: Reflection:
enabled: False
gaussian: Gaussian:
enabled: False
sigma: 1.5
gain: 2.0
iris: 22
red_balance: 1.709
```
#### New result:
```
--- # Settings
bidirectional: False
blue_balance: 1.081
brightness: 1.0
exposure_time: 0:00:00.008333
filters: 
    contrast: 
        enabled: True
        threshold: 5.0
    outlier: 
        enabled: True
        threshold: 5.0
    saturated: 
        enabled: True
    reflection: 
        enabled: False
    gaussian: 
        enabled: False
        sigma: 1.5
gain: 2.0
iris: 22
red_balance: 1.709
```

The above is valid YAML, so that one can easily construct a nested dictionary from the serialization:
```
import yaml
yaml.load(str(s))
```
```
{'bidirectional': False,
 'blue_balance': 1.081,
 'brightness': 1.0,
 'exposure_time': 0.008333,
 'filters': {'contrast': {'enabled': True, 'threshold': 5.0},
  'outlier': {'enabled': True, 'threshold': 5.0},
  'saturated': {'enabled': True},
  'reflection': {'enabled': False},
  'gaussian': {'enabled': False, 'sigma': 1.5}},
 'gain': 2.0,
 'iris': 22,
 'red_balance': 1.709}
```

### DIFF
--- a/modules/zivid/settings.py
+++ b/modules/zivid/settings.py
@@ -33,9 +33,9 @@ class Settings:  # pylint: disable=too-many-instance-attributes, too-few-public-
                 return False
 
             def __str__(self):
-                return """Contrast:
-enabled: {}
-threshold: {}""".format(
+                return """
+        enabled: {}
+        threshold: {}""".format(
                     self.enabled, self.threshold
                 )
 
@@ -63,9 +63,9 @@ threshold: {}""".format(
                 return False
 
             def __str__(self):
-                return """Outlier:
-enabled: {}
-threshold: {}""".format(
+                return """
+        enabled: {}
+        threshold: {}""".format(
                     self.enabled, self.threshold
                 )
 
@@ -89,8 +89,8 @@ threshold: {}""".format(
                 return False
 
             def __str__(self):
-                return """Saturated:
-enabled: {}""".format(
+                return """
+        enabled: {}""".format(
                     self.enabled
                 )
 
@@ -114,8 +114,8 @@ enabled: {}""".format(
                 return False
 
             def __str__(self):
-                return """Reflection:
-enabled: {}""".format(
+                return """
+        enabled: {}""".format(
                     self.enabled
                 )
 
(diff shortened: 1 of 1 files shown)


nedrebo:
Some background first.

1. A wrapper should not deal with this kind of stuff at all, it should be thin when possible and not re-implement basic stuff like toString. So the original code is already not correct, and you are patching something that should not even exist. Maintaining this is long term wasted work and in the meantime we have inconsistency between languages. Converting to native type and calling toString on it would be an inefficient(but not problem in real life) and more correct way of doing this as a quick fix. But not the ideal fix, more on that below.

2. Settings is only one of many data models that all should have identical interfaces. With this patch we get internal consistency as well. __str__ for a FrameInfo is still on the old format. All python code for all data models should be 100% auto generated from the C++ header. That is what we need to do to make this maintainable and consistent.

Some generic feedback. Printing the type should not print its type/name. `str(1)` is `'1'` not `'int: 1'`.

The prefixes should be applied by the code doing the formatting. In this case that is just moving the prefixes "one level out" in the code, giving the same result except no top level `---Settings:` which I think is a good thing. If people want to compose multiple settings into one file, they probably want to choose they way they do it them self.

nedrebo:
We have a vision for the Python wrapper and we know it is not ideal yet. There are multiple fundamental things missing. I would recommend that we set aside time and spend it on doing the proper fixes and not spend time on temporary patches it. Patches are welcome but before implementing thing I would recommend checking if there already are plans for how to proceed.

PR #24, opened 2019-11-06
Releaseable point cloud

trym-b:
Instead of transforming the point cloud directly to a numpy buffer, we
should instead return a intermediate layer, since in the underlying C++
API point cloud exposes various functions such as width and height.

Making buffer releasable also created some technical debt:
assertNotReleased is needed since accessing the underlying a C++ object
after it has been released causes a crash.

### DIFF
--- a/modules/zivid/point_cloud.py
+++ b/modules/zivid/point_cloud.py
@@ -1,0 +1,66 @@
+"""Contains the PointCloud class."""
+import numpy
+
+import _zivid
+
+
+class PointCloud:
+    """A point cloud."""
+
+    def __init__(self, internal_point_cloud):
+        """Create a point cloud from an internal point cloud.
+
+        Args:
+            internal_point_cloud: a internal point cloud
+
+        """
+        if not isinstance(internal_point_cloud, _zivid.PointCloud):
+            raise ValueError(
+                "Unsupported type for argument internal_point_cloud. Got {}, expected {}".format(
+                    type(internal_point_cloud), type(_zivid.PointCloud)
+                )
+            )
+        self.__impl = internal_point_cloud
+
+    def to_array(self):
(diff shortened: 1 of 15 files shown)


modules/zivid/point_cloud.py
```diff
@@ -0,0 +1,66 @@
...
+        Returns:
+            a numpy array
+
+        """
+        self.__impl.assert_not_released()
+        return numpy.array(self.__impl)
+
+    @property
+    def height(self):
+        """Maximum height of point cloud.
```
nedrebo: When 1:1 wrapping a function from C++ just copy the documentation.

 * Consistency
 * Time already spent on thinking about wording. No point in spending that time again.

This goes for all functions in PR and (also for the all functions in the entire wrapper, but that is not part of this PR).

trym-b: Fixed.

trym-b:
Ready for next round.

modules/zivid/point_cloud.py
```diff
@@ -0,0 +1,66 @@
...
+            a positive integer
+
+        """
+        return self.__impl.width()
+
+    def release(self):
+        """Release the underlying resources."""
+        self.__impl.release()
+
+    def __enter__(self):
```
nedrebo: Maybe we should have a mixin or base class or something for these three functions. They are duplicated many places now.

You are just adding to the existing solution, so no need to fix in this PR. But if you agree please add an issue for it.

trym-b: Could probably be done through a decorator:
```
@utils.context_manager
class SomethingThatRequiresContextManager:
    ....
```
Will create an issue

trym-b: https://github.com/zivid/zivid-python/issues/27

PR #28, opened 2019-11-13
2D API wrapper

trym-b:
Enables 2D API in this wrapper, so that you can easily capture 2D images
using a camera. 2D sample capture and tests has been included.

### DIFF
--- a/modules/zivid/__init__.py
+++ b/modules/zivid/__init__.py
@@ -12,7 +12,10 @@ from zivid.application import Application
 from zivid.camera import Camera
 from zivid.camera_state import CameraState
 from zivid.frame import Frame
+from zivid.frame_2d import Frame2D
 from zivid.frame_info import FrameInfo
+from zivid.image import Image
 from zivid.point_cloud import PointCloud
 from zivid.sdk_version import SDKVersion
 from zivid.settings import Settings
+from zivid.settings_2d import Settings2D
--- a/modules/zivid/camera.py
+++ b/modules/zivid/camera.py
@@ -1,6 +1,8 @@
 """Contains Camera class."""
 from zivid.frame import Frame
+from zivid.frame_2d import Frame2D
 import zivid._settings_converter as _settings_converter
+import zivid._settings_2d_converter as _settings_2d_converter
 import zivid._camera_state_converter as _camera_state_converter
 import _zivid
 
@@ -110,6 +112,24 @@ class Camera:
                 )
             )
         return Frame(self.__impl.capture())
+
+    def capture_2d(self, settings_2d):
+        """Capture a single 2D frame.
+
+        Note that the provided settings will only apply to this current 2D capture, and not future 3D captures.
+
+        Args:
+            settings_2d: Settings to use for the capture
+
+        Returns:
+            A frame containing a 2D image and metadata.
+
+        """
+        return Frame2D(
+            self.__impl.capture_2d(
+                _settings_2d_converter.to_internal_settings_2d(settings_2d)
+            )
+        )
 
     @property
     def state(self):
--- a/modules/zivid/frame_2d.py
+++ b/modules/zivid/frame_2d.py
@@ -1,0 +1,91 @@
+"""Contains the Frame class."""
+import _zivid
+
+import zivid._settings_2d_converter as _settings_converter
+import zivid._camera_state_converter as _camera_state_converter
+import zivid._frame_info_converter as _frame_info_converter
+from zivid.image import Image
+
+
+class Frame2D:
+    """A 2D frame captured by a Zivid camera.
+
+    Contains a 2D image as well as metadata, settings and state of the API at the time of capture.
+    """
+
+    def __init__(self, internal_frame_2d):
+        """Can only be initialized by an internal Zivid 2D frame.
+
+        Args:
+            internal_frame_2d: internal 2D frame
+
+        """
+        if isinstance(internal_frame_2d, _zivid.Frame2D):
+            self.__impl = internal_frame_2d
+        else:
+            raise TypeError(
+                "Unsupported type for argument internal_frame_2d. Got {}, expected {}.".format(
+                    type(internal_frame_2d).__name__, _zivid.Frame2D
+                )
+            )
+
+    def __str__(self):
+        return str(self.__impl)
+
+    def image(self):
+        """Return the underlying 2D image.
+
+        Returns:
+            an image instance
+
+        """
+        return Image(self.__impl.image())
+
+    @property
+    def settings(self):
+        """Get the settings 2d for the API at the time of the 2d frame capture.
+
+        Returns:
+            a settings 2d instance
+
+        """
+        return _settings_converter.to_settings_2d(  # pylint: disable=protected-access
+            self.__impl.settings
+        )
+
+    @property
+    def state(self):
+        """Get the camera state data at the time of the frame capture.
+
+        Returns:
+            a camera state instance
+
+        """
+        return _camera_state_converter.to_camera_state(  # pylint: disable=protected-access
+            self.__impl.state
+        )
+
+    @property
+    def info(self):
+        """Get information collected at the time of the frame capture.
+
+        Returns:
+            a camera info instance
+
+        """
+        return _frame_info_converter.to_info(  # pylint: disable=protected-access
+            self.__impl.info
+        )
+
+    def release(self):
+        """Release the underlying resources."""
+        self.__impl.release()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        self.release()
+
+    def __del__(self):
+        self.release()
--- a/modules/zivid/image.py
+++ b/modules/zivid/image.py
@@ -1,0 +1,78 @@
+"""Contains a the Image class."""
+import numpy
+
+import _zivid
+
+
+class Image:
+    """A 2-dimensional image stored on the host."""
+
+    def __init__(self, internal_image):
+        """Can only be initialized with an zivid internal image.
+
+        Args:
+            internal_image: an internal image
+
+        """
+        if not isinstance(internal_image, _zivid.Image):
+            raise ValueError(
+                "Unsupported type for argument internal_image. Got {}, expected {}".format(
+                    type(internal_image), type(_zivid.Image)
+                )
+            )
+        self.__impl = internal_image
+
+    def save(self, file_path):
+        """Save the image to a file.
+
+        The supported file type is PNG with extension .png.
+        This method will throw an exception if failing to save to the provided file_path.
+
+        Args:
+            file_path: destination path
+
+        """
+        self.__impl.save(str(file_path))
+
+    def to_array(self):
+        """Convert the image into a numpy array.
+
+        Returns:
+            a numpy array
+
+        """
+        self.__impl.assert_not_released()
+        return numpy.array(self.__impl)
+
+    @property
+    def height(self):
+        """Return height (number of rows) of image.
+
+        Returns:
+            an integer
+
+        """
+        return self.__impl.height()
+
+    @property
+    def width(self):
+        """Return width (number of columns) of image.
+
+        Returns:
+            an integer
+
+        """
+        return self.__impl.width()
+
+    def release(self):
+        """Release the underlying resources."""
+        self.__impl.release()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        self.release()
+
+    def __del__(self):
+        self.release()
--- a/src/ReleasableImage.cpp
+++ b/src/ReleasableImage.cpp
@@ -1,0 +1,36 @@
+#include <ZividPython/ReleasableImage.h>
+
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
+
+namespace
+{
+    py::buffer_info makeBufferInfo(ZividPython::ReleasableImage &image)
+    {
+        const auto data = image.dataPtr();
+
+        return py::buffer_info{
+            const_cast<Zivid::RGBA8 *>(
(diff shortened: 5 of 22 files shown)


modules/zivid/camera.py
```diff
@@ -111,6 +113,24 @@ def capture(self, settings_collection=None):
             )
         return Frame(self.__impl.capture())
 
+    def capture_2d(self, settings_2d):
```
eskaur: Camera.capture() works with no arguments, but Camera.capture_2d() does now. Is this intentional?

modules/zivid/frame_2d.py
```diff
@@ -0,0 +1,91 @@
+"""Contains the Frame class."""
+import _zivid
+
+import zivid._settings_2d_converter as _settings_converter
+import zivid._camera_state_converter as _camera_state_converter
+import zivid._frame_info_converter as _frame_info_converter
+from zivid.image import Image  # should simply be "from zivid import Image"
```
eskaur: Is this comment still relevant?

test/test_frame_2d.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+# @pytest.mark.physical_camera
+# def test_image_context_manager(physical_camera_frame_2d):
+#     import zivid
+#
+#     with physical_camera_frame_2d.image() as image:
+#         image = frame_2d.image
+#         assert image is not None
+#         assert isinstance(image, zivid.Frame2D)
```
eskaur: Can these commented lines be removed?

src/ReleasableImage.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+namespace py = pybind11;
+
+namespace
+{
+    py::buffer_info makeBufferInfo(ZividPython::ReleasableImage &image)
+    {
+        const auto data = image.dataPtr();
+
+        return py::buffer_info{ const_cast<void *>(static_cast<const void *>(data)), // HACKY HACK WATCH OUT
```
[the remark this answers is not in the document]
trym-b: We do not have a non const dataPtr for the 2d image class, so we need to hack around this by casting away to const. This should ideally be done in another way, perhaps @nedrebo has some ideas how to do this properly.

modules/zivid/frame_2d.py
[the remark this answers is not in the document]
trym-b: No, will delete it.

modules/zivid/camera.py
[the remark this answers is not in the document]
trym-b: That is a design decision that was taken when designing the 2d API for the c++ layer, I am simply trusting that this was designed properly.

src/ReleasableImage.cpp
```diff
@@ -0,0 +1,34 @@
...
+                                py::format_descriptor<Zivid::RGBA8>::format(),
+                                2,
+                                { image.height(), image.width() },
+                                { sizeof(Zivid::RGBA8) * image.width(), sizeof(Zivid::RGBA8) } };
+    }
+} // namespace
+
+namespace ZividPython
+{
+    //void wrapClass(pybind11::class_<Zivid::Image<Zivid::RGBA8>> pyClass)
```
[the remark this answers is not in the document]
trym-b: Will fix

src/Wrapper.cpp
```diff
@@ -21,13 +22,23 @@ ZIVID_PYTHON_MODULE // NOLINT
...
     ZIVID_PYTHON_WRAP_CLASS_AS_SINGLETON(module, Application);
     ZIVID_PYTHON_WRAP_CLASS_AS_RELEASABLE(module, Camera);
     ZIVID_PYTHON_WRAP_CLASS_AS_RELEASABLE(module, Frame);
+    ZIVID_PYTHON_WRAP_CLASS_AS_RELEASABLE(module, Frame2D);
+
+    //ZividPython::wrapClass<Zivid::Image<Zivid::RGBA8>, ZividPython::WrapType::normal>(
+    //    module,
+    //    static_cast<void (*)(pybind11::class_<Zivid::Image<Zivid::RGBA8>>)>(ZividPython::wrapClass),
+    //    "Image",
+    //    pybind11::buffer_protocol());
```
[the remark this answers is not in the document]
trym-b: Will fix

src/include/ZividPython/Wrappers.h
```diff
@@ -55,6 +55,10 @@ namespace ZividPython
     ZividPython::wrapClass<Zivid::name, ZividPython::WrapType::normal>(                                                \
         dest, static_cast<void (*)(pybind11::class_<Zivid::name>)>(ZividPython::wrapClass), #name)
 
+#define ZIVID_PYTHON_WRAP_STRUCT(dest, name)                                                                           \
```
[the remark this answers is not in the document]
trym-b: Old prototype functions from earlier, nice catch

test/test_frame_2d.py
trym-b: Re-added the test

trym-b:
Ready for next round

src/ReleasableImage.cpp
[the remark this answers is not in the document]
nedrebo: I don't think we need a hack here, but if we do we need a nice description about why.

However I would surprised if pybind11 did not support making numpy views out of const memory areas. Did you look for that? If not it is actual worth a bug report, but I would look thoroughly first, because this is something that I think is logical to have in this library.

modules/zivid/camera.py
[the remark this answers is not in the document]
eskaur: Sounds alright then.

src/ReleasableImage.cpp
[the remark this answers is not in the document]
eskaur: I'll hold back the Approve until it's either improved or the comment line is made more helpful/descriptive.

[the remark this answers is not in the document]
trym-b: Sounds good, will look around for a fix.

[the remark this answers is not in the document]
nedrebo: I looked at the source for pybind11, no support for const void *, but a flag for read only was added here: https://github.com/pybind/pybind11/pull/1466

I would use that and also change the cast to const_cast to be explicit about what is going on. Cast to void is not necessary, so this should be enough: `const_cast<PixelFormat *>(data)`.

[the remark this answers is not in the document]
trym-b: Updated the code and comment

src/ReleasableImage.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace py = pybind11;
+
+namespace
+{
+    py::buffer_info makeBufferInfo(ZividPython::ReleasableImage &image)
+    {
+        const auto data = image.dataPtr();
+
+        return py::buffer_info{ const_cast<Zivid::RGBA8 *>(data), // TODO: Const casting this until https://github.com/pybind/pybind11/pull/1466 has been merged
```
nedrebo: `const_cast` will still be needed after that PR.

https://github.com/pybind/pybind11/issues/1993 will make it possible to remove `const_cast`

Can you test what happens if you try to modify the numpy array and then for example call save on the original image? Do you save the changes or not or does it crash?

src/Wrapper.cpp
```diff
@@ -21,13 +22,18 @@ ZIVID_PYTHON_MODULE // NOLINT
...
     ZIVID_PYTHON_WRAP_DATA_MODEL(module, CameraState);
     ZIVID_PYTHON_WRAP_DATA_MODEL(module, FrameInfo);
     ZIVID_PYTHON_WRAP_DATA_MODEL(module, CameraIntrinsics);
 
     ZIVID_PYTHON_WRAP_CLASS_AS_SINGLETON(module, Application);
     ZIVID_PYTHON_WRAP_CLASS_AS_RELEASABLE(module, Camera);
     ZIVID_PYTHON_WRAP_CLASS_AS_RELEASABLE(module, Frame);
+    ZIVID_PYTHON_WRAP_CLASS_AS_RELEASABLE(module, Frame2D);
+
+    ZIVID_PYTHON_WRAP_CLASS_BUFFER_AS_RELEASABLE(module, Image);
```
nedrebo: Move down so we are in line with the original system of grouping the same macros together.

src/ReleasableImage.cpp
trym-b: Calling the following:

```
>>> import zivid
>>> app = zivid.Application()
>>> cam = app.connect_camera()
>>> fra2 = cam.capture_2d(zivid.Settings2d())
>>> im = fra2.image()
>>> im.save("orig.png")
>>> arr = im.to_array()
>>> for i in range(400,500):
...     for j in range (400, 500):
...         arr[i,j][1] = 0
...         arr[i,j][2] = 0
...         arr[i,j][3] = 0
... 
>>> im.save("new.png")
```

This gives me two images that looks exactly the same.

trym-b:
Ready for next round @nedrebo

modules/zivid/__init__.py
```diff
@@ -24,3 +24,8 @@
 from zivid.settings import (
     Settings,  # noqa: F401 'zivid.settings.Settings' imported but unused
 )
+from zivid.settings_2d import (
+    Settings2d,  # noqa: F401 'zivid.settings.Settings2d' imported but unused
```
nedrebo: Small d? Intentional?

modules/zivid/__init__.py
```diff
@@ -24,3 +24,8 @@
...
+)
+from zivid.image import Image  # noqa: F401 'zivid.image.Image' imported but unused
+from zivid.frame_2d import Frame2d  # noqa: F401 'zivid.frame_2d.d' imported but unused
```
nedrebo: Same

modules/zivid/image.py
```diff
@@ -0,0 +1,78 @@
+"""Contains a the Image class."""
+import numpy
+
+import _zivid
+
+
+class Image:
+    """A 2-dimensional image stored on the host."""
```
nedrebo: What is the host? I see this is just copied from C++, but I have no idea what it means. I would open an issue against C++ SDK if you agree.

src/include/ZividPython/Releasable.h
```diff
@@ -2,6 +2,12 @@
 
 #include <optional>
 
+#define ZIVID_PYTHON_FORWARD_0_ARGS_TEMPLATE_WRAP_RETURN(ReturnType, functionName, returnTypetypename)                 \
```
nedrebo: returnTypetypename sounds a bit weird. Looks like a type from its usage so it should have uppercase first letter, ReturnTypeTypename would be better.

I also se that there is already some inconsistency in this file. arg1type below is a type as well, ReturnType is also a type. The cleanest is probably to use camelCase for all macro args everything, and not CapiTaliZed at all.

An other thing is that I have a hard time understanding how ReturnTypeTypename is different from ReturnType. Maybe call it TempleteArg1 or templateArg1.

For concistency the macro should be called ZIVID_PYTHON_FORWARD_0_ARGS_TEMPLATE_1_ARG_WRAP_RETURN

modules/zivid/__init__.py
nedrebo: My proposal:
zivid.frame_2d: snake_case cant use D, keep as is
Frame2d: CapiTaliZed, D allowed we should use it.

modules/zivid/__init__.py
trym-b: It was intentional, but I do not have strong opinions about this, can change it to Settings2D.

modules/zivid/image.py
trym-b: Will create an issue

src/include/ZividPython/Releasable.h
trym-b: Ok, will change the name of the macro, but I will do the other cleanup in a separate PR: https://github.com/zivid/zivid-python/pull/32

src/ReleasableImage.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace
+{
+    py::buffer_info makeBufferInfo(ZividPython::ReleasableImage &image)
+    {
+        const auto data = image.dataPtr();
+
+        return py::buffer_info{
+            const_cast<Zivid::RGBA8 *>(
+                data), // TODO: Const casting this until https://github.com/pybind/pybind11/issues/1993 has been solved
```
nedrebo: This is now merged. The fix is now to simply remove the const cast once pybind11 is updated.

modules/zivid/frame_2d.py
```diff
@@ -0,0 +1,91 @@
...
+            raise TypeError(
+                "Unsupported type for argument internal_frame_2d. Got {}, expected {}.".format(
+                    type(internal_frame_2d).__name__, _zivid.Frame2D
+                )
+            )
+
+    def __str__(self):
+        return str(self.__impl)
+
+    def image(self):
```
apartridge: This wrapper does not seem to have the same pixelformat template/generic parameter that the C++ and .NET wrapper has. What if we add RGB8 or RGBA16, how will that be handled in the Python wrapper?

It could be that this method shuld have a parameter to specify the pixel format, or maybe it should be named imageRGBA8().  Then we can add other image* methods when we have more formats.

nedrebo: What we discussed earlier was to have only one function, but with a string parameter for datatype and then a data class where those strings can be found.

Multiple functions seems less nice.

src/ReleasableImage.cpp
trym-b: Updated the comment

modules/zivid/frame_2d.py
trym-b: We can easily extend this by adding a keyword-argument for pixel-format when we need it.

trym-b:
Any more comments @apartridge, @nedrebo or @nedrebo?

modules/zivid/frame_2d.py
apartridge: ok sounds good

PR #30, opened 2019-11-21
Hand-eye calibration support

trym-b:
This commit introduces 1.6.0's hand-eye calibration functionality,
allowing users to easily calibrate a camera with their robot.

Hand-eye calibration sample has been added.

### DIFF
--- a/.pylintrc-packaged-files
+++ b/.pylintrc-packaged-files
@@ -1,0 +1,10 @@
+[MESSAGES CONTROL]
+disable=missing-docstring,
+        duplicate-code, ### Temporarily disabled until sample files begin to look different
+        bad-continuation, ### Disabled because in conflict with black
+
+[FORMAT]
+max-line-length=120
+
+# Allow long lines if they contains long strings
+ignore-long-lines="[^"]{80,}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,9 @@ endif()
 
 include(${CMAKE_BINARY_DIR}/conan.cmake)
 
-conan_cmake_run(REQUIRES pybind11/2.3.0@conan/stable
+conan_cmake_run(REQUIRES 
+                    eigen/3.3.7@conan/stable
+                    pybind11/2.3.0@conan/stable
                 BASIC_SETUP)
 
 set(CMAKE_CXX_STANDARD 17)
@@ -24,5 +26,6 @@ endif()
 
 find_package(Zivid ${ZIVID_SDK_VERSION} EXACT COMPONENTS Core REQUIRED)
 find_package(pybind11 2.3.0 REQUIRED)
+find_package(Eigen3 3.3 REQUIRED)
 
 add_subdirectory(src)
--- a/modules/zivid/handeye.py
+++ b/modules/zivid/handeye.py
@@ -1,0 +1,224 @@
+"""Contains hand-eye calibration functions and classes."""
+import _zivid
+
+
+def detect_feature_points(point_cloud):
+    """Detect feature points from a calibration object in a point cloud.
+
+    The functionality is to be exclusively used in combination with Zivid verified checkerboards.
+    For further information please visit Zivid help page.
+
+    Args:
+        point_cloud: cloud from a frame containing an image of a calibration object
+
+    Returns:
+        Instance of DetectionResult
+
+    """
+    return DetectionResult(
+        _zivid.handeye.detect_feature_points(
+            point_cloud._PointCloud__impl  # pylint: disable=protected-access
+        )
+    )
+
+
+def calibrate_eye_in_hand(calibration_inputs):
+    """Perform eye-in-hand calibration.
+
+    The procedure requires feature point sets acquired at the minimum from two poses.
+    All the input poses have to be different. The feature point sets cannot be empty.
+    All the feature point sets have to have same number of feature points.
+    An exception will be thrown if the above requirements are not fulfilled.
+
+    Args:
+        calibration_inputs: a Sequence of CalibrationInput instances
+
+    Returns:
+        Instance of CalibrationOutput
+
+    """
+    return CalibrationOutput(
+        _zivid.handeye.calibrate_eye_in_hand(
+            [
+                calib._CalibrationInput__impl  # pylint: disable=protected-access
+                for calib in calibration_inputs
+            ]
+        )
+    )
+
+
+def calibrate_eye_to_hand(calibration_inputs):
+    """Perform eye-to-hand calibration.
+
+    The procedure requires feature point sets acquired at the minimum from two poses.
+    All the input poses have to be different. The feature point sets cannot be empty.
+    All the feature points have to have same number of elements.
+    An exception will be thrown if the above requirements are not fulfilled.
+
+    Args:
+        calibration_inputs: a Sequence of CalibrationInput instances
+
+    Returns:
+        Instance of CalibrationOutput
+
+    """
+    return CalibrationOutput(
+        _zivid.handeye.calibrate_eye_to_hand(
+            [
+                calib._CalibrationInput__impl  # pylint: disable=protected-access
+                for calib in calibration_inputs
+            ]
+        )
+    )
+
+
+class Pose:  # pylint: disable=too-few-public-methods
+    """Describes a robot pose."""
+
+    def __init__(self, matrix):
+        """Pose constructor taking a 4x4 transform.
+
+        The constructor throws if the input transform does not describe pure rotation and translation.
+
+        Args:
+            matrix: a 4x4 numpy array
+
+        """
+        self.__impl = _zivid.handeye.Pose(matrix)
+
+    def __str__(self):
+        return self.__impl.to_string()
+
+
+class DetectionResult:  # pylint: disable=too-few-public-methods
+    """A result returned by the detect_feature_points(...) call."""
+
+    def __init__(self, impl):
+        """Initialize from internal DetectionResult.
+
+        Args:
+            impl: an internal DetectionResult Instance
+
+        """
+        self.__impl = impl
+
+    def __str__(self):
+        return self.__impl.to_string()
+
+    def __bool__(self):
+        return self.__impl.valid()
+
+
+class CalibrationInput:  # pylint: disable=too-few-public-methods
+    """Binds together a robot pose and the detection result acquired from the pose."""
+
+    def __init__(self, pose, detected_features):
+        """Binds together a robot pose and the detection result acquired from the pose.
+
+        Args:
+            pose: a robot pose
+            detected_features: a DetectionResult instance
+
+        """
+        self.__impl = _zivid.handeye.CalibrationInput(
+            pose._Pose__impl,  # pylint: disable=protected-access
+            detected_features._DetectionResult__impl,  # pylint: disable=protected-access
+        )
+
+    def __str__(self):
+        return self.__impl.to_string()
+
+
+class CalibrationOutput:
+    """The calibration result containing the computed pose and reprojection errors for all the input poses."""
+
+    def __init__(self, internal_calibration_result):
+        """Calibration results containing the computed pose and reprojection errors for all the input poses.
+
+        Args:
+            internal_calibration_result: An internal zivid calibration result instance
+
+        """
+        self.__impl = internal_calibration_result
+
+    @property
+    def per_pose_calibration_residuals(self):
+        """Return the calibration residuals.
+
+        Feature points (for each input pose) are transformed into a common frame.
+        A rigid transform between feature points and corresponding centroids are utilized to
+        compute residuals for rotational and translational parts. An exception is thrown if the result is not valid.
+
+        Returns:
+            a list of calibration residuals
+
+        """
+        return [
+            CalibrationResidual(element)
+            for element in self.__impl.perPoseCalibrationResiduals()
+        ]
+
+    @property
+    def hand_eye_transform(self):
+        """Hand-eye transform.
+
+        A 4x4 matrix describing hand-eye calibration transform (either eye-in-hand or eye-to-hand).
+        An exception is thrown if the result is not valid.
+
+        Returns:
+            A 4x4 numpy matrix
+
+        """
+        return self.__impl.handEyeTransform()
+
+    @property
+    def valid(self):
+        """Test if CalibrationOutput is valid.
+
+        Returns:
+            a bool
+
+        """
+        return self.__impl.valid()
+
+    def __bool__(self):
+        return self.valid
+
+    def __str__(self):
+        return self.__impl.to_string()
+
+
+class CalibrationResidual:
+    """Binds together rotational and translational residual errors wrt. calibration transform."""
+
+    def __init__(self, internal_calibration_residual):
+        """Binds together rotational and translational residual errors wrt. calibration transform.
+
+        Args:
+            internal_calibration_residual: an internal calibration residual instance
+
+        """
+        self.__impl = internal_calibration_residual
+
+    @property
+    def rotation(self):
+        """Component of calibration residual.
+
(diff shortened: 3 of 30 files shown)


modules/zivid/handeye.py
```diff
@@ -0,0 +1,122 @@
...
+        Args:
+            matrix: a 4x4 numpy array
+
+        """
+        self.__impl = _zivid.handeye.Pose(matrix)
+
+    def __str__(self):
+        return self.__impl.to_string()
+
+
```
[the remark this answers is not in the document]
trym-b: Added a commit that fixes this.

modules/zivid/handeye.py
```diff
@@ -0,0 +1,122 @@
...
+        """Initialize from internal DetectionResult.
+
+        Args:
+            impl: an internal DetectionResult Instance
+
+        """
+        self.__impl = impl
+
+    def __str__(self):
+        return self.__impl.to_string()
```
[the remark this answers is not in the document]
trym-b: This is not what I get as output. If I run `print(result)` I get: `Valid: 0`

trym-b:
Added unit tests

modules/zivid/handeye.py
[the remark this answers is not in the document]
nedrebo: `Valid: 0` is also suspicious. I would expect `Valid: yes` or `Valid: no`. I hope that is the output you saw from C++ @eskaur.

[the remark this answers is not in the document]
eskaur: Unfortunately no, @nedrebo . 
Success in the C++ sample gives:
> Result:{ Valid: 1, Detected feature points: {.......<bunch of vector3s>........} }
Failure gives: 
> Result:{ Valid: 0, Detected feature points: {  } }

Better than what I got from the Python in this branch, but still not perfect.

modules/zivid/handeye.py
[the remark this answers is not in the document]
eskaur: Tested. Seems to work.

modules/zivid/handeye.py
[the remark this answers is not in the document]
eskaur: Turns out the missing "1" or "0" in str(detection_result) is only an issue in Jupyter Notebook... Works as expected in regular python interpreter.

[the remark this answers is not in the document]
nedrebo: Ah, the detection result is not a data model, that is why you don't get "yes/no".

The issue here is that we should use `std::boolalpha` in `Zivid::Serialization::toString`. This is an issue in the C++ interface and I think it warrens an issue. "0"/"1" is not very readable.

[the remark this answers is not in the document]
nedrebo: Created: ZIVID-2420

trym-b:
Any more comments on this one? @eskaur  or @nedrebo ?

nedrebo:
> Any more comments on this one? @eskaur or @nedrebo ?

I will juts quickly look at it once @eskaur is done, I have not looked at it yet.

samples/sample_calibrate_eye_to_hand.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def _enter_robot_pose(index):
+    inputted = input(
+        "Enter pose with id={} (a line with 16 space separated values describing 4x4 row-major matrix):".format(
+            index
+        )
+    )
+    elements = inputted.split(maxsplit=15)
+    data = np.array(elements, dtype=np.float64).reshape((4, 4)).transpose()
```
eskaur: The user shouldn't have to transpose the numpy matrix before passing it into our API. We should rather make the wrapper so that it just works, given a matrix that is correct in numpy.

I'll look into this now.

eskaur: Ah yes, Eigen defaults to column-major storage, while we want row-major (both the numpy matrix and our internal matrix is row-major).

I'll put the necessary changes in comments on the appropriate lines.

test/test_hand_eye.py
```diff
@@ -0,0 +1,36 @@
+def test_init_pose():
+    import numpy as np
+    import zivid.handeye
+
+    elements = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]
+    data = np.array(elements, dtype=np.float64).reshape((4, 4)).transpose()
```
eskaur: Remove this transpose() once the wrapper is set to row-major.

test/test_hand_eye.py
```diff
@@ -0,0 +1,36 @@
...
+
+def test_calibration_input_init_failure(point_cloud):
+    import numpy as np
+    import pytest
+    import zivid.handeye
+
+    feature_points = zivid.handeye.detect_feature_points(point_cloud)
+
+    elements = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]
+    data = np.array(elements, dtype=np.float64).reshape((4, 4)).transpose()
```
eskaur: Remove this transpose() once the wrapper is set to row-major.

src/Pose.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
+
+namespace ZividPython
+{
+    void wrapClass(pybind11::class_<Zivid::HandEye::Pose> pyClass)
+    {
+        pyClass.def(py::init([](const Eigen::Matrix4d &matrix) {
```
eskaur: Change this line to: 
`pyClass.def(py::init([](const Eigen::Matrix<double, 4, 4, Eigen::RowMajor> &matrix) `

src/include/ZividPython/Matrix.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "pybind11/eigen.h"
+#include <Zivid/Matrix.h>
+
+namespace ZividPython::Conversion
+{
+    template<typename T, int rows, int cols>
+    auto toCpp(const Eigen::Matrix<T, rows, cols> &source)
```
eskaur: Change this line to:
`auto toCpp(const Eigen::Matrix<T, rows, cols, Eigen::RowMajor> &source)`

src/include/ZividPython/Matrix.h
```diff
@@ -0,0 +1,19 @@
...
+    {
+        return Zivid::Matrix<T, rows, cols>{ source.data(), source.data() + source.size() };
+    }
+
+    template<typename T, size_t rows, size_t cols>
+    auto toPy(const Zivid::Matrix<T, rows, cols> &source)
+    {
+        return Eigen::Matrix<T, rows, cols>{ &(source(0, 0)) };
```
eskaur: Change this line to:
`return Eigen::Matrix<T, rows, cols, Eigen::RowMajor>{ &(source(0, 0)) };`

samples/sample_calibrate_eye_to_hand.py
```diff
@@ -0,0 +1,77 @@
...
+        elif command == "c":
+            calibrate = True
+        else:
+            print("Unknown command '{}'".format(command))
+
+    print("Performing hand-eye calibration...")
+    calibration_result = zivid.handeye.calibrate_eye_to_hand(calibration_inputs)
+    if calibration_result:
+        print("OK")
+        print("Result:\n{}".format(calibration_result))
```
eskaur: Converting calibration_result to string doesn't give expected behavior. The actual numbers for the residuals are missing, and we get stuff like this:
`
{ Hand-Eye transformation: [ [ 0.670959,  0.677084, -0.302277,  637.027719], 
  [ 0.735697, -0.557012,  0.385340, -598.974910], 
  [ 0.092536, -0.480932, -0.871861,  1159.374538], 
  [ 0.000000,  0.000000,  0.000000,  1.000000] ], Per-pose calibration residuals: { { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  }, { Residual for rotation in deg: , Residual for translation in mm:  } } }
`

eskaur: Also, trying to run `calibration_result.perPoseCalibrationResiduals()` gives
`TypeError: Unable to convert function return value to a Python type! The signature was
	(self: _zivid._zivid.handeye.CalibrationOutput) -> std::vector<Zivid::HandEye::CalibrationResidual, std::allocator<Zivid::HandEye::CalibrationResidual> >

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.
`

modules/zivid/handeye.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    return _zivid.handeye.calibrate_eye_in_hand(
+        [
+            calib._CalibrationInput__impl  # pylint: disable=protected-access
+            for calib in calibration_inputs
+        ]
+    )
+
+
+def calibrate_eye_to_hand(calibration_inputs):
```
eskaur: I've locally recreated the corresponding C++ CLI-tool for this one (the one that takes real poses and pictures from files), and confirmed that the resulting transform is the same given the same input-data.

modules/zivid/handeye.py
```diff
@@ -0,0 +1,125 @@
...
+
+    """
+    return DetectionResult(
+        _zivid.handeye.detect_feature_points(
+            point_cloud._PointCloud__impl  # pylint: disable=protected-access
+        )
+    )
+
+
+def calibrate_eye_in_hand(calibration_inputs):
```
eskaur: This one doesn't actually work, because you have forgotten to add it to `HandEye.cpp`

trym-b:
Ready for next round @eskaur

modules/zivid/handeye.py
kontramind: Can we expect to have this CLI to ship it to the Customer?

eskaur: We should definitely consider it.

eskaur:
Since the eye-in-hand now also works, I compared this one with the C++ CLI-tool as well. It all seems to work fine. :)

modules/zivid/handeye.py
eskaur: It works now, since the last fixups.

test/test_hand_eye.py
```diff
@@ -0,0 +1,36 @@
...
+    assert pose is not None
+    assert isinstance(pose, zivid.handeye.Pose)
+
+
+def test_detect_feature_points(point_cloud):
+    import zivid
+
+    feature_points = zivid.handeye.detect_feature_points(point_cloud)
+
+    assert feature_points is not None
```
eskaur: Is it worth noting that this test actually fails at detecting feature-point (since MiscObjects.zdf does not contain a checkerboard). Could be misleading...

CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
 project(zivid LANGUAGES CXX)
-
```
nedrebo: Keep newline

CMakeLists.txt
```diff
@@ -9,9 +8,10 @@ endif()
 
 include(${CMAKE_BINARY_DIR}/conan.cmake)
 
-conan_cmake_run(REQUIRES pybind11/2.3.0@conan/stable
+conan_cmake_run(REQUIRES 
+                    eigen/3.3.7@conan/stable
+                    pybind11/2.3.0@conan/stable
                 BASIC_SETUP)
-
```
nedrebo: Keep newline

eskaur:
Besides my comment about one of the unit-tests, this looks good to me now.
I'll leave it to @nedrebo to do final approval.

test/test_hand_eye.py
trym-b: Perhaps, but simply taking a random point cloud, how likely is it that this pointcloud contains a checkerboard? If it really was a checkerboard in that point cloud, I would have named it `point_cloud_with_checkerboard` or somethin like that

trym-b:
Ready for next round @nedrebo

src/include/ZividPython/Wrappers.h
```diff
@@ -52,8 +54,14 @@ namespace ZividPython
...
 
 #define ZIVID_PYTHON_WRAP_CLASS(dest, name)                                                                            \
-    ZividPython::wrapClass<Zivid::name, ZividPython::WrapType::normal>(                                                \
-        dest, static_cast<void (*)(pybind11::class_<Zivid::name>)>(ZividPython::wrapClass), #name)
+    ZividPython::wrapClass<name, ZividPython::WrapType::normal>(dest,                                                  \
+                                                                static_cast<void (*)(pybind11::class_<name>)>(         \
+                                                                    ZividPython::wrapClass),                           \
+                                                                #name)
+
+#define ZIVID_PYTHON_WRAP_CLASS_FULLPATH(dest, name, full_path)                                                        \
```
nedrebo: Delete this

src/HandEye.cpp
```diff
@@ -0,0 +1,41 @@
...
+        ZIVID_PYTHON_WRAP_CLASS(dest, CalibrationOutput);
+        ZIVID_PYTHON_WRAP_CLASS(dest, CalibrationInput);
+        ZIVID_PYTHON_WRAP_CLASS(dest, DetectionResult);
+        ZIVID_PYTHON_WRAP_CLASS(dest, CalibrationResidual);
+
+        dest.def("detect_feature_points",
+                 [](const ReleasablePointCloud &releasablePointCloud) {
+                     return Zivid::HandEye::detectFeaturePoints(releasablePointCloud.impl());
+                 })
+            .def("calibrate_eye_in_hand",
```
nedrebo: I do not think you need a lambda for the two last ones.

src/ReleasableCamera.cpp
```diff
@@ -14,7 +14,7 @@ namespace ZividPython
 {
     void wrapClass(pybind11::class_<ReleasableCamera> pyClass)
     {
-        pyClass.def(py::init<>())
+        pyClass.def(py::init())
             .def(py::self == py::self) // NOLINT
```
nedrebo: Im am not sure if there are any comparison operators in the wrapped C++ code, but you should check if it is and add operator for those?

Side note: Also are we sure all of these operators are actually in the python layer as well. Should check that, but fixing that is an other issue.

.pylintrc-packaged-files
```diff
@@ -0,0 +1,10 @@
+[MESSAGES CONTROL]
```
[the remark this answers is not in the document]
trym-b: Perhaps, not entirely sure about that, since it is a rather dangerous thing to do.

src/ReleasableCamera.cpp
trym-b: We are not using any of the equality operators in the `zivid` module, will create an issue for it

PR #39, opened 2019-12-03
Adding Capture Assistant support

kontramind:
This is initial commit. The commit adds:
1. AmbientLightFrequency
2. SuggestSettingsParammeters
3. suggestSettings
4. HDR::capture

### DIFF
--- a/modules/zivid/captureassistant.py
+++ b/modules/zivid/captureassistant.py
@@ -1,0 +1,82 @@
+"""Contains capture assistant functions and classes."""
+import _zivid
+import zivid._settings_converter as _settings_converter
+from zivid._make_enum_wrapper import _make_enum_wrapper
+
+
+AmbientLightFrequency = _make_enum_wrapper(
+    _zivid.captureassistant.AmbientLightFrequency,
+    "Ensure compatibility with the frequency of the ambient light in the scene.",
+)
+
+
+class SuggestSettingsParameters:  # pylint: disable=too-few-public-methods
+    """Input to the Capture Assistant algorithm.
+
+    Used to specify a constraint on the total capture time for the settings suggested by the Capture Assistant,
+    and optionally specify the ambient light frequency.
+    The capture time constraint assumes a computer meeting Zivid's recommended minimum compute power.
+
+    """
+
+    def __init__(self, max_capture_time, ambient_light_frequency=None):
+        """Initialize SuggestSettingsParameters.
+
+        Args:
+            max_capture_time: an instance of datetime.timedelta
+            ambient_light_frequency: a member of the enum zivid.captureassistant.AmbientLightFrequency
+
+        """
+        if ambient_light_frequency is None:
+            self.__impl = _zivid.captureassistant.SuggestSettingsParameters(
+                max_capture_time
+            )
+        else:
+            self.__impl = _zivid.captureassistant.SuggestSettingsParameters(
+                max_capture_time,
+                ambient_light_frequency._to_internal(),  # pylint: disable=protected-access
+            )
+
+    @property
+    def max_capture_time(self):
+        """Get max capture time.
+
+        Returns:
+            Instance of datetime.timedelta
+
+        """
+        return self.__impl.maxCaptureTime()
+
+    @property
+    def ambient_light_frequency(self):
+        """Get ambient light frequency.
+
+        Returns:
+            Instance of AmbientLightFrequency
+
+        """
+        return AmbientLightFrequency(self.__impl.ambientLightFrequency().name)
+
+    def __str__(self):
+        return self.__impl.to_string()
+
+
+def suggest_settings(camera, suggest_settings_parameters):
+    """Find settings for the current scene based on the suggest_settings_parameters.
+
+    The suggested settings returned from this function should be passed into hdr.capture to perform the actual capture.
+
+    Args:
+        camera: an instance of zivid.Camera
+        suggest_settings_parameters: an instance of zivid.captureassistant.SuggestSettingsParameters which provides
+                                     parameters (e.g., max capture time constraint) to the suggest_settings algorithm.
+
+    Returns:
+        List of Settings.
+
+    """
+    internal_settings = _zivid.captureassistant.suggest_settings(
+        camera._Camera__impl,  # pylint: disable=protected-access
+        suggest_settings_parameters._SuggestSettingsParameters__impl,  # pylint: disable=protected-access
+    )
+    return [_settings_converter.to_settings(internal) for internal in internal_settings]
--- a/modules/zivid/hdr.py
+++ b/modules/zivid/hdr.py
@@ -1,6 +1,7 @@
 """Contains a global HDR method."""
 import _zivid
 from zivid.frame import Frame
+import zivid._settings_converter as _settings_converter
 
 
 def combine_frames(frames):
@@ -19,5 +20,27 @@ def combine_frames(frames):
     return Frame(
         _zivid.hdr.combine_frames(
             [frame._Frame__impl for frame in frames]  # pylint: disable=protected-access
+        )
+    )
+
+
+def capture(camera, settings_list):
+    """Capture and merge frames.
+
+    Args:
+        camera: an instance of zivid.Camera
+        settings_list: a list of zivid.Settings instances
+
+    Returns:
+        a single frame (when single setting given) or high dynamic range frame (when multiple settings given)
+
+    """
+    return Frame(
+        _zivid.hdr.capture(
+            camera._Camera__impl,  # pylint: disable=protected-access
+            [
+                _settings_converter.to_internal_settings(settings)
+                for settings in settings_list
+            ],
         )
     )
--- a/src/SuggestSettingsParameters.cpp
+++ b/src/SuggestSettingsParameters.cpp
@@ -1,0 +1,19 @@
+#include <ZividPython/CaptureAssistant.h>
+#include <ZividPython/SuggestSettingsParameters.h>
+
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
+
+namespace ZividPython
+{
+    void wrapClass(pybind11::class_<Zivid::CaptureAssistant::SuggestSettingsParameters> pyClass)
+    {
+        pyClass.def(py::init<std::chrono::milliseconds>(), py::arg("max_capture_time"))
+            .def(py::init<std::chrono::milliseconds, Zivid::CaptureAssistant::AmbientLightFrequency>(),
+                 py::arg("max_capture_time"),
+                 py::arg("ambient_light_frequency"))
+            .def("maxCaptureTime", &Zivid::CaptureAssistant::SuggestSettingsParameters::maxCaptureTime)
+            .def("ambientLightFrequency", &Zivid::CaptureAssistant::SuggestSettingsParameters::ambientLightFrequency);
+    }
+} // namespace ZividPython
--- a/src/Wrapper.cpp
+++ b/src/Wrapper.cpp
@@ -3,7 +3,9 @@
 #include <ZividPython/DataModelWrapper.h>
 #include <ZividPython/Wrappers.h>
 
+#include <ZividPython/AmbientLightFrequency.h>
 #include <ZividPython/CameraRevision.h>
+#include <ZividPython/CaptureAssistant.h>
 #include <ZividPython/Environment.h>
 #include <ZividPython/Firmware.h>
 #include <ZividPython/HDR.h>
@@ -14,6 +16,7 @@
 #include <ZividPython/ReleasableFrame2D.h>
 #include <ZividPython/ReleasablePointCloud.h>
 #include <ZividPython/SingletonApplication.h>
+#include <ZividPython/SuggestSettingsParameters.h>
 #include <ZividPython/Version.h>
 #include <ZividPython/Wrapper.h>
 
@@ -46,4 +49,5 @@ ZIVID_PYTHON_MODULE // NOLINT
     ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, HDR);
     ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, Version);
     ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, HandEye);
+    ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, CaptureAssistant);
 }
--- a/src/include/ZividPython/Wrappers.h
+++ b/src/include/ZividPython/Wrappers.h
@@ -17,7 +17,7 @@ namespace ZividPython
     {
         normal,
         releasable,
-        singleton
+        singleton,
     };
 
     template<typename Source, WrapType wrapType, typename WrapFunction, typename... Tags>
@@ -41,6 +41,13 @@ namespace ZividPython
         wrapFunction(pyClass);
     }
 
+    template<typename Source, typename WrapFunction>
+    void wrapEnum(const pybind11::module &dest, const WrapFunction &wrapFunction, const char *exposedName)
+    {
+        auto pyEnum = pybind11::enum_<Source>{ dest, exposedName, pybind11::dynamic_attr() };
+        wrapFunction(pyEnum);
+    }
+
     template<typename WrapFunction>
     void wrapNamespaceAsSubmodule(pybind11::module &dest,
                                   const WrapFunction &wrapFunction,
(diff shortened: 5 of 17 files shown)


modules/zivid/captureassistant.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class AmbientLightFrequency(Enum):  # pylint: disable=too-few-public-methods
+    """Ensure compatibility with the frequency of the ambient light in the scene."""
+
+    hz50 = _zivid.captureassistant.AmbientLightFrequency.hz50
+    hz60 = _zivid.captureassistant.AmbientLightFrequency.hz60
+    none = _zivid.captureassistant.AmbientLightFrequency.none
+
+    def __str__(self):
```
eskaur: Why overload `__str__()`? 
The default implementation from `Enum` would yield e.g. `AmbientLightFrequency.hz50`, which is perhaps even more readable.

kontramind: This could be true. ATM, rationale was to be close to c++ implementation.
This could surely be modified.
@nedrebo your take?

modules/zivid/captureassistant.py
```diff
@@ -0,0 +1,85 @@
...
+    The capture time constraint assumes a computer meeting Zivid's recommended minimum compute power.
+
+    """
+
+    def __init__(self, budget, frequency=None):
+        """Initialize SuggestSettingsParameters.
+
+        Args:
+            budget: max capture time
+            frequency: ambient light frequency
```
eskaur: Should we say more here about type? 
- "Budget" seems to accept a number (interpreted as seconds) or datetime.timedelta.
- "Frequency" only accepts members ("versions?") of AmbientLightFrequency enum. This documentation alone does not hint the user about the existence of this enum.

modules/zivid/captureassistant.py
```diff
@@ -0,0 +1,85 @@
...
+        return self.__impl.to_string()
+
+
+def suggest_settings(camera, suggest_settings_parameters):
+    """Find settings for the current scene based on the suggest_settings_parameters.
+
+    The suggested settings returned from this function should be passed into hdr.capture to perform the actual capture.
+
+    Args:
+        camera: reference to Camera instances
```
eskaur: Two comments:
- I don't thing we should use the word "reference" here. All mutable variables in Python are references, so it's a given.
- Should be singular, right? "Camera instance"
So perhaps just do:

> camera      Camera instance

kontramind: Agree.

modules/zivid/captureassistant.py
kontramind: Think also that budget -> max_capture_time

modules/zivid/hdr.py
```diff
@@ -21,3 +22,25 @@ def combine_frames(frames):
             [frame._Frame__impl for frame in frames]  # pylint: disable=protected-access
         )
     )
+
+
+def capture(camera, settings_list):
+    """Capture and merge frames.
+
+    Args:
+        camera: a reference to camera instance
```
eskaur: > camera:  Camera instance
(see other comment above)

samples/sample_capture_assistant.py
```diff
@@ -0,0 +1,25 @@
...
+    suggest_settings_parameters = SuggestSettingsParameters(
+        budget=datetime.timedelta(milliseconds=1200),
+        frequency=AmbientLightFrequency.hz50,
+    )
+
+    suggested_settings = zivid.captureassistant.suggest_settings(
+        camera, suggest_settings_parameters
+    )
+
+    frame = zivid.hdr.capture(camera, suggested_settings)
```
eskaur: You should use the context manager idiom, like in the other samples (see `sample_capture_hdr.py`).
It doesn't really matter in a simple example like this, but we should do it for consistency.

```
with zivid.hdr.capture(camera, suggested_settings) as as hdr_frame:
    hdr_frame.save("result.zdf"
```

modules/zivid/captureassistant.py
```diff
@@ -0,0 +1,85 @@
+"""Contains capture assistant functions and classes."""
```
eskaur: As discussed IRL, this module should probably be called `capture_assistant`. This would reflect the convention of `point_cloud` and such.

test/test_captureassistant.py
```diff
@@ -0,0 +1,54 @@
...
+    assert str(AmbientLightFrequency.none) == "none"
+
+
+def test_suggest_settings_parameters():
+    from zivid.captureassistant import AmbientLightFrequency, SuggestSettingsParameters
+
+    suggest_settings_parameters = SuggestSettingsParameters(
+        budget=datetime.timedelta(milliseconds=1200),
+        frequency=AmbientLightFrequency.hz50,
+    )
```
eskaur: We seem to automatically support using a float directly as `budget`, which is then interpreted as seconds.
Should we officially support that? If so, we should test for that.

modules/zivid/captureassistant.py
nedrebo: Yes, this is just tmp code to make sure the rest could be implemented and tested. How to wrap enums are still undecided. We have the standard requirements about robust code, but I did not spend time on actually coming up with a design. As it is stands now we are leaking pybind11 types `zivid`, that is not something we can merge. pybind11 should not be observable from `zivid` (using "normal" code).

modules/zivid/hdr.py
```diff
@@ -1,6 +1,7 @@
 """Contains a global HDR method."""
 import _zivid
 from zivid.frame import Frame
+import zivid._settings_converter as _settings_converter
```
nedrebo: I don't think you need to do this manually, did you try to use the zivid.Settings ctor and passsing in _zivid.Settings?

modules/zivid/captureassistant.py
eskaur: In which sense are we leaking pybind11 types?
Is it because
```
hzmode = zivid.captureassistant.AmbientLightFrequency.hz50
print(type(hzmode.value))
```
gives 
> <class '_zivid._zivid.captureassistant.AmbientLightFrequency'>
?

eskaur: We could perhaps do the following:

```
# Hidden mapping to underlying pybind11 type
_AmbientFrequencyChoices = ["hz50", "hz60", "none"]
_AmbientFrequencyMap = {choice: getattr(_zivid.captureassistant.AmbientLightFrequency, choice) 
                        for choice in _AmbientFrequencyChoices}

AmbientLightFrequency = enum.Enum('AmbientLightFrequency', _AmbientFrequencyChoices)

# Visible enum with normal integer underlying type
AmbientLightFrequency = enum.Enum('AmbientLightFrequency', [*_AmbientFrequencyChoices])

class SuggestSettingsParameters: 
    def __init__(self, budget, frequency=None):
        self.__impl = _zivid.captureassistant.SuggestSettingsParameters(
            budget, _AmbientFrequencyMap[frequency.name]
        )

```

This has the following features: 
- completely hides the pybind11 type in the private/"underscore"  dictionary.
- defines the mapping to the pybind11 type in a single place.
- Impossible to mistakenly map e.g. "hz50" to "hz60"

modules/zivid/hdr.py
kontramind: Actually, here it is used to go from public ones to internal ones, see the line below:
` _settings_converter.to_internal_settings(settings)`
there seems to no ctor which would allow _zivid.Settings(zivid.Settings).

nedrebo: You are correct. Keep it as is. I see the same is done in `zivid.Camera.settings`.

modules/zivid/captureassistant.py
kontramind: Have changed doc strings with fixup. 
There I'm more precise that budget (now max_capture_time) expects an instance of datetime.timedelta. Have addressed frequency (now ambient_light_frequency) in similar manner.

The acceptance of floats for budget may be a side-effect (feature?) of pybind11 - not sure if some action is needed at this point.

modules/zivid/hdr.py
kontramind: This has been also addressed by fixup.

modules/zivid/captureassistant.py
kontramind: Addressed in fixup.

modules/zivid/captureassistant.py
kontramind: Done in fixup

samples/sample_capture_assistant.py
kontramind: Done in fixup.

modules/zivid/captureassistant.py
kontramind: Done in fixup.

modules/zivid/captureassistant.py
eskaur: I think it's fine like this. We don't need to officially support using a float for budget.

eskaur: Could `an instance of zivid.capture_assistant.AmbientLightFrequency` be misleading? It implies that it is a class that you would instantiate with a constructor like
```
my_instance = zivid.captureassistant.AmbientLightFrequency()
```
I think it should say:
`a value of the enum zivid.capture_assistant.AmbientLightFrequency`

modules/zivid/captureassistant.py
kontramind: As discussed IRL, this could be a good candidate (bit changed @eskaur 's suggestion)
`_ALFEntries = {entry: entry for entry in _zivid.capture_assistant.AmbientLightFrequency.__entries}
_ALFToInternalMap = {entry: getattr(_zivid.capture_assistant.AmbientLightFrequency, entry) for entry in _ALFEntries}

AmbientLightFrequency = Enum('AmbientLightFrequency', _ALFEntries)
setattr(AmbientLightFrequency, '__str__', lambda self: str(self.name))`

modules/zivid/captureassistant.py
kontramind: done in fixup

modules/zivid/captureassistant.py
kontramind: now in fixup

eskaur: Looks good to me, but I'll let @nedrebo decide if it properly addressed his initial concern.

kontramind: There are some linter issues - will try to fix them

kontramind:
ready for next round

src/Wrapper.cpp
```diff
@@ -46,4 +49,5 @@ ZIVID_PYTHON_MODULE // NOLINT
     ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, HDR);
     ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, Version);
     ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, HandEye);
+    ZIVID_PYTHON_WRAP_NAMESPACE_AS_SUBMODULE(module, Capture_Assistant);
```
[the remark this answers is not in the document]
kontramind: As agreed, have flipped back to no underscore - zivid.captureassistant.
At least we are consistent now with handeye nomenclature.
Preferred naming should take place in a separate PR.

src/include/ZividPython/Wrappers.h
```diff
@@ -17,7 +17,8 @@ namespace ZividPython
     {
         normal,
         releasable,
-        singleton
+        singleton,
+        enum_class,
```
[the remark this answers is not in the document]
kontramind: Indeed. Addressed in a fixup.

src/SuggestSettingsParameters.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
+
+namespace ZividPython
+{
+    void wrapClass(pybind11::class_<Zivid::CaptureAssistant::SuggestSettingsParameters> pyClass)
+    {
+        pyClass
+            .def(py::init([](std::chrono::milliseconds &max_capture_time) {
```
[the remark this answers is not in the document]
kontramind: Thanks. Corrected in fixup.

modules/zivid/capture_assistant.py
```diff
@@ -0,0 +1,96 @@
...
+
+_AmbientLightFrequencyEntries = {  # pylint: disable=invalid-name
+    entry: entry
+    for entry in _zivid.capture_assistant.AmbientLightFrequency.__entries  # pylint: disable=protected-access
+}
+_AmbientLightFrequencyToInternalMap = {  # pylint: disable=invalid-name
+    entry: getattr(_zivid.capture_assistant.AmbientLightFrequency, entry)
+    for entry in _AmbientLightFrequencyEntries
+}
+AmbientLightFrequency = Enum(  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
kontramind: Simple hard-coded and checked implementation may be:
```
class AmbientLightFrequency(Enum):
    hz50 = "hz50"
    hz60 = "hz60"
    none = "none"

    def __str__(self):
        return str(self.name)


assert AmbientLightFrequency.__members__.keys() == _zivid.captureassistant.AmbientLightFrequency.__members__.keys(), f"{AmbientLightFrequency} not in sync with C++ implementation"

```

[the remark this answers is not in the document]
kontramind: The above may be improved (e.g. to avoid hz50 = "hz60"):

```
class StrEnumMeta(EnumMeta):
    def __new__(metacls, cls, bases, oldclassdict):
        newclassdict = _EnumDict()
        for k, v in oldclassdict.items():
            if v == ():
                v = k
            newclassdict[k] = v
        return super().__new__(metacls, cls, bases, newclassdict)

class AutoStrEnum(str, Enum, metaclass=StrEnumMeta):
    "base class for name=value str enums"

class AmbientLightFrequency(AutoStrEnum):
    hz50 = ()
    hz60 = ()
    none = ()

    def __str__(self):
        return str(self.name)

assert AmbientLightFrequency.__members__.keys() == _zivid.captureassistant.AmbientLightFrequency.__members__.keys(), f"{AmbientLightFrequency} not in sync with C++ implementation"
```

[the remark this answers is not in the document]
kontramind: Another implementation:

- automatically in sync with values exposed by pybind11

- but, could not make __str__ work without dynamically adding the method inlater step

- unsure if auto-completion works everywhere (does in IPython)

´´´
class _AmbientLightFrequencyEnumMeta(EnumMeta):

    def __new__(metacls, cls, bases, oldclassdict):
        newclassdict = _EnumDict()
        for member in _zivid.captureassistant.AmbientLightFrequency.__members__:
            newclassdict[member] = str(member)
        return super().__new__(metacls, cls, bases, newclassdict)


class AmbientLightFrequency(str, Enum, metaclass=_AmbientLightFrequencyEnumMeta):
    "base class for name=value str enums"


setattr(AmbientLightFrequency, "__str__", lambda self: str(self.name))
´´´

kontramind:
@nedrebo Together with @eskaur , metaclass based proposal was improved to provide a generic solution for exposing enums. 

Ready for next round.

modules/zivid/captureassistant.py
```diff
@@ -0,0 +1,82 @@
...
+import _zivid
+import zivid._settings_converter as _settings_converter
+from zivid._make_enum_wrapper import _make_enum_wrapper
+
+
+AmbientLightFrequency = _make_enum_wrapper(
+    _zivid.captureassistant.AmbientLightFrequency,
+    "Ensure compatibility with the frequency of the ambient light in the scene.",
+)
```
eskaur: Very nice, very clean :)
Auto-completion of options also seems to work everywhere.

eskaur:
Looks very good to me now. Glad we took the time to generalize the enum-wrapper.

You should rebase and squash.

kontramind:
re-based and squashed

modules/zivid/captureassistant.py
nedrebo: Nice :man_dancing:

kontramind:
@nedrebo @eskaur 
Maybe this PR can be approved? 
Or there are some more issues/concerns  to be addressed?

modules/zivid/captureassistant.py
```diff
@@ -0,0 +1,82 @@
+"""Contains capture assistant functions and classes."""
+import _zivid
+import zivid._settings_converter as _settings_converter
+from zivid._make_enum_wrapper import _make_enum_wrapper
+
+
+AmbientLightFrequency = _make_enum_wrapper(
```
trym-b: If this function call fails we will get exceptions when calling `import zivid`

kontramind: Is this different to other parts of our code which may fail?
Any suggestions to make it more robust?

trym-b: Can we not call this where we use this?

eskaur: It's needed when the captureassistant module is imported, so the autocomplete will work. As far as I can see, all the submodules are imported when zivid is imported.

PR #46, opened 2019-12-13
Improve error message when module build does not work

nedrebo:
This improve error messages during `pip install zivid` on systems
without binary packages and without C++17 support. This is implemented
as a nested exceptions where the outer most excetion gives a fix-it
hint.

I also sorted some dependency lists.

**Before**
before

**After**
after

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,15 +28,77 @@ def _zivid_python_version():
     return version
 
 
+def _make_message_box(*message):
+    width = max([len(e) for e in message])
+
+    box_bar = "+-" + "-" * width + "-+"
+    empty_line = "\n| " + " " * width + " |\n"
+    message_lines = ["| " + line + " " * (width - len(line)) + " |" for line in message]
+
+    return (
+        "\n\n" + box_bar + "\n" + empty_line.join(message_lines) + "\n" + box_bar + "\n"
+    )
+
+
 def _check_dependency(module_name, package_hint=None):
     if package_hint is None:
         package_hint = module_name
     if module_name not in [module[1] for module in iter_modules()]:
         raise ImportError(
-            "Missing module '{}'. Please install '{}' manually or use PIP>=19 to handle build dependencies automatically (PEP 517).".format(
-                module_name, package_hint
+            _make_message_box(
+                "!! Missing module '{}' !!".format(module_name),
+                "Please install '{}' manually or use PIP>=19 to handle build dependencies automatically (PEP 517)".format(
+                    package_hint
+                ),
             )
         )
+
+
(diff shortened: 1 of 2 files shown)


eskaur:
I'll take a look at this once the CI passes.

nedrebo:
> I'll take a look at this once the CI passes.

@eskaur: Ready

setup.py
```diff
@@ -28,26 +28,85 @@ def _zivid_python_version():
...
         package_hint = module_name
     if module_name not in [module[1] for module in iter_modules()]:
         raise ImportError(
-            "Missing module '{}'. Please install '{}' manually or use PIP>=19 to handle build dependencies automatically (PEP 517).".format(
-                module_name, package_hint
+            _make_message_box(
+                "!! Missing module '{}' !!".format(module_name),
+                "Please install '{}' manually or use PIP>=19 to handle build dependencies automatically (PEP 517).".format(
+                    package_hint
+                ),
```
eskaur: The new message box will obviously not appear if the missing module happens to be tabulate itself, but I guess that's alright?

setup.py
```diff
@@ -28,26 +28,85 @@ def _zivid_python_version():
     return version
 
 
+def _make_message_box(*message_lines):
+    from tabulate import tabulate
+
+    return "\n\n" + tabulate([["\n\n".join(message_lines)]], tablefmt="grid") + "\n"
+
+
```
eskaur: Looks very good though. They can't miss the error now :)

setup.py
```diff
@@ -28,26 +28,85 @@ def _zivid_python_version():
...
+            _make_message_box(
+                "!! Missing module '{}' !!".format(module_name),
+                "Please install '{}' manually or use PIP>=19 to handle build dependencies automatically (PEP 517).".format(
+                    package_hint
+                ),
             )
         )
 
 
+def _check_cpp17_compiler():
```
eskaur: Doesn't seem to work. I deleted GCC and tried to build. 
Got the old and obscure error message.

eskaur: Seems to work fine in a Docker. Ignore this for now :)

setup.py
nedrebo: Oh no, the error message will not even show, only an import error.

Fixed

eskaur:
Works great now. 

Tested on clean Ubuntu 16.04:
1) Nice error message when missing python modules.
2) Installed missing modules
3) Nice error message about missing C++17 compiler
4) Followed instructions in README to set up C++17
5) Builds without error

nedrebo:
Removed tabulate dependency @eskaur

PR #48, opened 2019-12-19
Snake case naming and rename module names

trym-b:
Makes the module names consistent


trym-b:
Will merge the two commits before final merge, just wanted to split it up for reviewr's pleasure.

src/include/ZividPython/Wrappers.h
```diff
@@ -13,6 +16,57 @@
 
 namespace ZividPython
 {
+    namespace
+    {
+        bool isCapital(char x)
```
kontramind: Use std::isupper()

kontramind:
Also, seems that sample_captureassistant.py should be renamed to sample_capture_assistant.py to keep it consistent.

modules/zivid/hand_eye.py
```diff
@@ -169,7 +169,7 @@ def hand_eye_transform(self):
             A 4x4 numpy matrix
 
         """
-        return self.__impl.handEyeTransform()
+        return self.__impl.hand_eyeTransform()
```
nedrebo: Functions should be snake_case. (Not camel_snakeCase :))

src/include/ZividPython/Wrappers.h
```diff
@@ -13,6 +16,49 @@
...
+                throw std::runtime_error{ "First character of string: '" + capitalized + "' is not capitalized" };
+            }
+            std::stringstream ss;
+            ss << char(tolower(capitalized[0]));
+
+            for(std::string::size_type i = 1; i < capitalized.size(); ++i)
+            {
+                if(isupper(capitalized[i]))
+                {
+                    std::string::size_type previous = i - 1;
```
nedrebo: I would have used auto for readability.

src/include/ZividPython/Wrappers.h
```diff
@@ -13,6 +16,49 @@
...
+                    std::string::size_type next = i + 1;
+
+                    if(isupper(capitalized[previous]) && (next < capitalized.size() && isupper(capitalized[next]))
+                       || (next >= capitalized.size()))
+                    {
+                        ss << char(tolower(
+                            capitalized[i])); // if surrounded by capital case (looking at the B character): ABC -> abc
```
nedrebo: It is nice to have the comment before the code. Else you read the code, figure out what it does, then when you reach the comment it is redundant.

Also trailing comment makes formatting weird.

I would put the comment all the way before the if, because that is actually what you are describing.

modules/zivid/hand_eye.py
trym-b: Fixed. This change was not intended. This was the result of a search-replace that was not case-sensitive

trym-b:
Ready for next round.

src/include/ZividPython/Wrappers.h
```diff
@@ -13,6 +16,50 @@
 
 namespace ZividPython
 {
+    namespace
+    {
+        std::string toSnakeCase(const std::string capitalized)
+        {
+            if(capitalized.empty())
```
kontramind: Wonder if we should also check for spaces (leading, between words,trailing) or any other characters like +.- ....?

src/include/ZividPython/Wrappers.h
```diff
@@ -13,6 +16,50 @@
...
+
+            for(auto i = 1; i < capitalized.size(); ++i)
+            {
+                if(isupper(capitalized[i]))
+                {
+                    auto previous = i - 1;
+                    auto next = i + 1;
+
+                    // if surrounded by capital case (looking at the B character): ABC -> abc
+                    if(isupper(capitalized[previous]) && (next < capitalized.size() && isupper(capitalized[next]))
```
kontramind: Should we consider using c++ regex as an more advanced approach?

src/include/ZividPython/Wrappers.h
trym-b: As long as this function is in this namespace and is not exposed anywhere else, I think we don't need to have tests for all kinds of special characters.

src/include/ZividPython/Wrappers.h
trym-b: I am open for alternatives, but I think this code is easy to understand and modify.

kontramind: Agree. The code is clear.
I was more thinking in general ... if we need more complex handling in future.
Let's not over-engineer :).

src/include/ZividPython/Wrappers.h
```diff
@@ -13,6 +16,50 @@
 
 namespace ZividPython
 {
+    namespace
+    {
+        std::string toSnakeCase(const std::string capitalized)
```
eskaur: If the variable name is supposed to describe the casing style, perhaps it should be changed from `capitalized` to `upperCamelCase` (which unfortunately itself has to be written in lower camel case ;) )
https://en.wikipedia.org/wiki/Camel_case

src/include/ZividPython/Wrappers.h
```diff
@@ -13,6 +16,50 @@
...
+            ss << char(tolower(capitalized[0]));
+
+            for(auto i = 1; i < capitalized.size(); ++i)
+            {
+                if(isupper(capitalized[i]))
+                {
+                    auto previous = i - 1;
+                    auto next = i + 1;
+
+                    // if surrounded by capital case (looking at the B character): ABC -> abc
```
eskaur: This comment is incomplete, since the following if-statement also introduces special handling if the final letter is capitalized, regardless of whether it is surrounded by capital case.

The current code yields:
`CamelCaseA` -> `camel_casea`
Is that intended behavior?

trym-b: No. or condition was in the wrong scope, when I moved it to the correct location it works in that test case.

eskaur: Tested new version. Looks good.

src/include/ZividPython/Wrappers.h
trym-b: Should be fixed now

PR #53, opened 2020-03-11
Bump Zivid SDK version to 1.8.1

nedrebo:


eskaur:
The top of the README still says that we require `version 1.8.0 or higher`, but that is not accurate. We require exactly one version if `pip install zivid` is the main way to set up.

Having 1.8.0 installed and doing `pip install zivid` leads to:
```
    Could not find a configuration file for package "Zivid" that exactly
    matches requested version "1.8.1".
  
    The following configuration files were considered but not accepted:
  
      /usr/lib/cmake/Zivid/ZividConfig.cmake, version: 1.8.0
```

eskaur:
Of course, if the README in the repo is supposed to apply to more advanced usage than just using `pip`, then it's another discussion.

nedrebo:
That was my thinking, to build the code you need 1.8.0 or newer. The requirements for `pip install` can be found here: https://github.com/zivid/zivid-python#pypi

I agree this can be confusing. Maybe we can improve the Readme overall in an other PR?

nedrebo:
Ready for next round. Probably best to review per commit.

PR #60, opened 2020-03-15
Change exposure values in sample_capture_2d.py

### LINKED ISSUE #57: Modify sample_capture_2d.py
https://github.com/zivid/zivid-python/blob/master/samples/sample_capture_2d.py#L10-L12
Add brightness and gain like in C++:
https://github.com/zivid/zivid-sdk/blob/bf89ba7cce266c1fa02a97fa26cd2d72b2620888/cpp/Samples/CPP/SampleCapture2D.cpp#L16-L19
Also, use the same parameter values.

SatjaSivcev:
Make the python sample consistent with the equivalent C++ sample

Fixes #57

### DIFF
--- a/samples/sample_capture_2d.py
+++ b/samples/sample_capture_2d.py
@@ -8,8 +8,9 @@ def _main():
     camera = app.connect_camera()
 
     settings_2d = zivid.Settings2D()
-    settings_2d.iris = 50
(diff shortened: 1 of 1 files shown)


SatjaSivcev:
@eskaur I've rebased and need approval.

PR #63, opened 2020-03-15
Change variable name in sample_capture_assistant,py

SatjaSivcev:
Make variable name consistent across python samples, in this case with sample_capture_hdr.py
This helps to simplify tutorials as they link to code snippets of different samples.

### DIFF
--- a/samples/sample_capture_assistant.py
+++ b/samples/sample_capture_assistant.py
@@ -13,11 +13,11 @@ def _main():
         ambient_light_frequency=AmbientLightFrequency.hz50,
     )
 
-    suggested_settings = zivid.capture_assistant.suggest_settings(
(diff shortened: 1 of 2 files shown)


eskaur:
I agree that we should be consistent where possible. But, maybe we should just go with `settings_list` across the Python samples instead? It is technically a list, and "collection" is a little formal.

SatjaSivcev:
> 
> 
> I agree that we should be consistent where possible. But, maybe we should just go with `settings_list` across the Python samples instead? It is technically a list, and "collection" is a little formal.

I agree, 'settings_list' sounds better and is technically correct. I've made changes now in both sample_capture_hdr and sample_capture_assistant.

Ready for the next round.

PR #67, opened 2020-04-17
Remove most of the hardcoded versions from pip requirements files

trym-b:
If we do not freeze the version we will automatically get the latest and
greatest of linting tools, formatting tools, etc. We would need to
update this repo whenever any linter tools get a new test we cannot
pass, but I think this is a worthwhile trade.

Build requirements should be frozen, so that building the code should
work. Test and lint requirements does not need to be frozen.

The newer pip modules required some changes:
    - pytest-unmarked was replaced with calling pytest with -m
    - added Raises section in docstrings
    - ignoring pytest import-outside-toplevel
    - change error type to TypeError when initializing camera, etc


nedrebo:
I have not yet looked at the refactoring, but it is probably good. It is also good to bump versions in tool chain, that should be done as often as we can. But we should keep the versions pinned, else the CI may break at any point in time for any user just because a new version of random python packages was released. This is a common issue and it happens all the time in python projects (where you typically rely on many modules from different sources) that don't pin all packages.

trym-b:
> I have not yet looked at the refactoring, but it is probably good. It is also good to bump versions in tool chain, that should be done as often as we can. But we should keep the versions pinned, else the CI may break at any point in time for any user just because a new version of random python packages was released. This is a common issue and it happens all the time in python projects (where you typically rely on many modules from different sources) that don't pin all packages.

That is true, but I also want to keep our python code up to date and not stagnate. Last time we updated was almost a year ago, it seems to me that we are not good at updating them.

nedrebo:
Then we must do better. We can try to set up https://requires.io/ with auto PRs and possible with badge also from https://shields.io/.

trym-b:
Closing this PR, creating a separate PR to update the packages to the latest versions.

PR #69, opened 2020-06-10
Fix broken CI tests with updated Intel OpenCL CPU driver

trym-b:
The CI tests have been failing frequently with the following error:

> An OpenCL error occurred: Failed to get devices for context.
> ...
> [CL_INVALID_CONTEXT]

After some digging into this using clinfo, I discovered that device
name like "Intel(R) Xeon(R) Platinum 8171M CPU @ 2.60GHz" fails, while
"Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz" works. This suggests that
our current Intel OpenCL CPU driver does not support all the CPU's
available in Azure Pipelines.

Updating the Intel driver from version 16.1.1 to 18.1.0 fixes this
problem.

The prior method of manually installing the OpenCL CPU driver based on
the .tgz archive caused problems (clinfo would just hang). From this
commit we are instead using the install.sh script provided by Intel.

We now also print out "clinfo" after installing the driver. This will
help troubleshooting issues like these in the future.

Note that the OpenCL CPU driver officially support Ubuntu 16.04 and
not 18.04 or 20.04. However, the release notes states that "higher
versions are expected to be functional", and my own testing suggest
this works fine. We must set "CONTINUE_WITH_OPTIONAL_ERROR=yes" in
the .cfg file to allow non-supported OS'es.

This commit was based on apartridge's commit:
https://github.com/zivid/zivid-ros/commit/e951b7ce4aaeb5d2ce9a429f2f55afa86a8a7f31

### DIFF
--- a/continuous-integration/linux/platform-dependent/fedora-30/setup.sh
+++ b/continuous-integration/linux/platform-dependent/fedora-30/setup.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 
+SCRIPT_DIR="$(realpath $(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd) )"
+
 dnf --assumeyes install \
     bsdtar \
+    clinfo \
     g++ \
     libatomic \
     python3-devel \
@@ -12,16 +15,12 @@ dnf --assumeyes install \
 alternatives --install /usr/bin/python python /usr/bin/python3 0 || exit $?
 alternatives --install /usr/bin/pip pip /usr/bin/pip3 0 || exit $?
 
-function install_opencl_cpu_runtime {
-    TMP_DIR=$(mktemp --tmpdir --directory zivid-setup-opencl-cpu-XXXX) || exit $?
-    pushd $TMP_DIR || exit $?
-    wget -nv https://www.dropbox.com/s/0cvg8fypylgal2m/opencl_runtime_16.1.1_x64_ubuntu_6.4.0.25.tgz || exit $?
-    tar -xf opencl_runtime_16.1.1_x64_ubuntu_6.4.0.25.tgz || exit $?
-    dnf install --assumeyes opencl_runtime_*/rpm/*.rpm || exit $?
-    popd || exit $?
-    rm -r $TMP_DIR || exit $?
-}
-
+source "$SCRIPT_DIR/../common.sh" || exit $?
+# Install OpenCL CPU runtime driver prerequisites
+dnf --assumeyes install \
+    numactl-libs \
+    redhat-lsb-core \
+    || exit $?
 install_opencl_cpu_runtime || exit $?
 
 function install_www_deb {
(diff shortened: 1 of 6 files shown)


continuous-integration/linux/platform-dependent/fedora-30/setup.sh
```diff
@@ -13,17 +14,46 @@ alternatives --install /usr/bin/python python /usr/bin/python3 0 || exit $?
...
+    echo "This warning can be ignored."
+    echo
+    ./install.sh --silent installer_config.cfg || exit $?
     popd || exit $?
     rm -r $TMP_DIR || exit $?
 }
 
 install_opencl_cpu_runtime || exit $?
 
+echo "clinfo:"
```
nedrebo, line 54: I would move this to the end of `Code/zivid/python/continuous-integration/linux/setup.sh` so that we guarantee that clinfo is always installed and executed.

continuous-integration/linux/platform-dependent/ubuntu-18.04/setup.sh
```diff
@@ -22,19 +23,43 @@ update-alternatives --install /usr/bin/python python /usr/bin/python3 0 || exit
 update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 0 || exit $?
 
 function install_opencl_cpu_runtime {
```
apartridge, line 25: Should move the part that installs the intel driver to a common place instead of duplicating it?

continuous-integration/linux/platform-dependent/fedora-30/setup.sh
```diff
@@ -13,17 +14,46 @@ alternatives --install /usr/bin/python python /usr/bin/python3 0 || exit $?
...
-    tar -xf opencl_runtime_16.1.1_x64_ubuntu_6.4.0.25.tgz || exit $?
-    dnf install --assumeyes opencl_runtime_*/rpm/*.rpm || exit $?
+    wget -q https://www.dropbox.com/s/h0txd04aqfluglq/l_opencl_p_18.1.0.015.tgz || exit $?
+    tar -xf l_opencl_p_18.1.0.015.tgz || exit $?
+    cd l_opencl_*/ || exit $?
+
+    cat > installer_config.cfg <<EOF
+# See silent.cfg in the .tgz for description of the options.
+ACCEPT_EULA=accept
+# 'yes' below is required because the installer officially supports only Ubuntu 16.04. However, it will
```
apartridge, line 33: Should fix this comment to make sense for fedora and arch also

trym-b: Fixed

continuous-integration/linux/platform-dependent/ubuntu-18.04/setup.sh
trym-b: Moved

trym-b:
Ready for next round

PR #70, opened 2020-06-10
Pin arch AUR zivid package version and fix broken OpenCL test setup

trym-b:
Since updating zivid AUR packages, the current master does no longer
build on arch. Pinning the version makes arch build again.
This is required to get zivid-python to pass CI before the wrapper is
updated to work with API 2.0.

Also updated OpenCL setup in order not to fail when run in the CI.

### DIFF
--- a/src/include/ZividPython/Releasable.h
+++ b/src/include/ZividPython/Releasable.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <optional>
+#include <stdexcept>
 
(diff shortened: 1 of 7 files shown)


src/include/ZividPython/Releasable.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <optional>
+#include <stdexcept>
```
nedrebo, line 4: This change does not seem to be related. Can it be removed?

PR #74, opened 2020-06-24
Restructure requirements files

trym-b:
Prework for updating the pip packages. The latest version of pylint
requires that it knows about various plugins for pytest in order to
successfully lint. This commit makes it easier to install the packages
for a specific task, so it is now easier to add the test packages before
linting.


azure-pipelines.yml
```diff
@@ -22,10 +22,10 @@ jobs:
       Python39:
         python.version: '3.9'
   steps:
-  - script: pip install -r continuous-integration\requirements-build-and-test.txt"
+  - script: pip install --requirement continuous-integration\python-requirements\build.txt
     displayName: 'Install python requirements'
   - script: python continuous-integration\windows\build.py --root %CD%
-    displayName: 'Build and test'
+    displayName: 'Build'
```
nedrebo, line 28: The label does not match script names in the matrix below.

azure-pipelines.yml
```diff
@@ -22,10 +22,10 @@ jobs:
       Python39:
         python.version: '3.9'
   steps:
-  - script: pip install -r continuous-integration\requirements-build-and-test.txt"
+  - script: pip install --requirement continuous-integration\python-requirements\build.txt
```
nedrebo, line 25: The requirement needed here is not "build". Here we are outside docker and "build" happens inside.

Maybe split the file in `ci.txt` and `build.txt`.

trym-b:
Discussed this IRL, the separation between setup, build, lint and test is poor. Will do some more changes before I come back with a more extensive PR.

PR #76, opened 2020-06-25
Restructure pipeline workflow

trym-b:
Prework for updating the pip packages. The latest version of pylint
requires that it knows about various plugins for pytest in order to
successfully lint. This demand requires the dependencies of setup, test
and lint to be separated, since before this commit all of the steps were
more or less done in one big blob.

By restructuring the azure-pipeline worklfow into setup, build and
test/lint it is much easier to reason about the pipeline, and now easier
to install the correct dependency for the correct stage.


continuous-integration/linux/test.sh
```diff
@@ -2,14 +2,8 @@
...
 ROOT_DIR=$(realpath "$SCRIPT_DIR/../..")
-VENV=$(mktemp --tmpdir --directory zivid-python-test-venv-XXXX) || exit $?
 
-python -m venv $VENV || exit $?
-source $VENV/bin/activate || exit $?
-
-pip install --upgrade pip || exit $?
-pip install "$ROOT_DIR" || exit $?
-pip install -r "$SCRIPT_DIR/../requirements-build-and-test.txt" || exit $?
+pip install --requirement "$SCRIPT_DIR/../python-requirements/test.txt" || exit $?
```
nedrebo, line 6: Why did you remove the virtual environment?

nedrebo: Now you will have access to packages installed by earlier steps.

trym-b: I decided to remove the virtual environments because I do not think they provide much value as of now. We could move the virtual creation to the build.sh step, but I decided against it, mostly because it will be difficult to reason about the code in test.sh (just by looking at that file it is not clear that the python/pip we use, which might be confusing).

PR #79, opened 2020-06-25
Move sample tests to pytest

trym-b:
Instead of running a special test just for running samples, we should
instead use pytest to handle this.


test/test_samples.py
```diff
@@ -0,0 +1,48 @@
...
+    current_working_directory = Path(os.getcwd()).resolve()
+    sample = (
+        current_working_directory
+        / ".."
+        / ".."
+        / "samples"
+        / "sample_print_version_info.py"
+    ).resolve()
+
+    subprocess.check_output(args=("python", str(sample),))
```
nedrebo, line 48: The big block in the middle looks like it is duplicated from above except for a minor difference that is probably a copy-paste error. Can this be made into a reusable function?

```suggestion
def test_capture_from_file(sample_data_file):
    run_sample(
        name="capture_from_file",
        working_directory=sample_data_file.parent
    )

def test_print_version_info():
    run_sample(
        name="print_version_info"
    )
```

trym-b:
Thanks for reviewing @eskaur and @nedrebo

PR #81, opened 2020-07-14
Enable Windows CI tests

trym-b:
In order to run tests on Windows, Intel OpenCL runtime needs to be
installed to run the tests on a compute device. It is also required to
create a configuration file so that the tests are run on the CPU instead
of a GPU (since there are no GPU available in azure-pipelines)

Added update environment script in order to load the latest envrionement
variables set by `Zivid-SDK` after successful installation. `PATH`
environment variable must be updated befre running the tests.

Ignored sample testing for windows, because of the following error:
```
python: can't open file 'D:\a\samples\sample_print_version_info.py': [Errno 2] No such file or directory
```

Ignored `winreg` import error, this module is not available on Linux

### DIFF
--- a/continuous-integration/windows/setup.py
+++ b/continuous-integration/windows/setup.py
@@ -1,6 +1,7 @@
 import tempfile
 import sys
 import subprocess
+import shutil
 from pathlib import Path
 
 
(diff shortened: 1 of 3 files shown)


eskaur:
Little typo in commit message: "envrionement"

continuous-integration/windows/setup.py
```diff
@@ -42,9 +42,51 @@ def _install_zivid_sdk():
...
+    CrashHandler: yes
+    Directory: ""
+    Enabled: yes
+    FlushOnExit: yes
+    HistorySize: 30
+    MaxFileSizeMB: 50
+    MaxTotalSizeMB: 1000
+    RotationCount: 2
+    WriteInterval: 0
+"""
```
eskaur, line 79: Can't you use "ZividAPIConfig.yml" like we do in the Linux setup?

test/test_samples.py
```diff
@@ -1,11 +1,21 @@
+import platform
+
 import pytest
 
 
+@pytest.mark.skipif(
+    platform.system() == "Windows",
+    reason=r"python: can't open file 'D:\a\samples\sample_capture_from_file.py': [Errno 2] No such file or directory",
```
eskaur, line 8: Maybe you can make an issue to investigate this further later.

continuous-integration/windows/setup.py
trym-b: Good point, did not think about this

test/test_samples.py
trym-b: https://github.com/zivid/zivid-python/issues/82

trym-b:
Ready for next round @eskaur

trym-b:
Thank you for reviewing @eskaur

test/test_samples.py
nedrebo: It is hard to find the link between this code and #82.
 * #82 should mention that these two tests should be re-enabled.
 * The `reason` here should reference the #82.

PR #83, opened 2020-07-14
Avoid mismatch in Python version

torbsorb:
It is possible that calling pip directly does not install packages for the Python accessed with Python directly. To avoid mismatch, call pip via Python.
`python -m pip install -r requirements.txt`
vs
`pip install -r requirements.txt`

### DIFF
--- a/continuous-integration/windows/test.py
+++ b/continuous-integration/windows/test.py
@@ -48,6 +48,8 @@ def _install_pip_dependencies():
     print("Installing python test requirements", flush=True)
     _run_process(
         (
+            "python",
+            "-m",
             "pip",
             "install",
             "--requirement",


trym-b:
How is this possible? Why would your python point to some version 1.2.3 while your pip points to 2.3.4?

torbsorb:
I tried to fix it, but pip points to my 3.8 installation, while python points to 3.7. It shouldn't be that difficult, but then again it's quite easy to just call via python.. I believe many of the responses I found recommended just that.

torbsorb:
Mainly this thread that convinced me to not bother trying to fix it: https://stackoverflow.com/questions/2812520/dealing-with-multiple-python-versions-and-pip

trym-b:
Ok, sounds good then

trym-b:
But you should probably put this into the commit message.

PR #87, opened 2020-08-31
Fix errors from __del__ methods

eskaur:
Currently, if e.g. a PointCloud gets scheduled for deletion before
self.__impl has been set (e.g. due to an error in __init__),
an AttributeError would be thrown from __del__. Exceptions are
ignored when coming from __del__, but may still cause an error
message to be written to stderr, which looks bad.

The problem can be illustrated by running:
```
pytest -c ./pytest.ini test/test_point_cloud.py --capture=no -k test_illegal_init
```

This commit solves this by checking that self.__impl exists before
calling release() on it.

### DIFF
--- a/modules/zivid/application.py
+++ b/modules/zivid/application.py
@@ -67,7 +67,12 @@ class Application:
 
     def release(self):
         """Release the underlying resources."""
-        self.__impl.release()
+        try:
+            impl = self.__impl
+        except AttributeError:
+            pass
+        else:
+            impl.release()
 
     def __enter__(self):
         return self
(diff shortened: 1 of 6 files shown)


nedrebo (approved):
Approved, but seel like this releasable concept now is so much code to that we probably should have a mixin or similar. Something like `class Application(Releasable(_zivid.Application)):` that just do the right thing.

eskaur:
> Approved, but seems like this releasable concept now is so much code to that we probably should have a mixin or similar. Something like `class Application(Releasable(_zivid.Application)):` that just do the right thing.

I agree; I was uncomfortable with this degree of code repetition already. I didn't want to restructure this at this point though.

Issue #91, opened 2020-09-23
toSnakeCase could be simplified


anordal:
This should be exactly equivalent:

```diff
--- a/src/include/ZividPython/Wrappers.h
+++ b/src/include/ZividPython/Wrappers.h
@@ -18,6 +18,8 @@ namespace ZividPython
 {
     namespace
     {
+        /// Inserts underscore on positive case flank and before negative case flank:
+        /// ZividSDKVersion -> zivid_sdk_version
         std::string toSnakeCase(const std::string upperCamelCase)
         {
             if(upperCamelCase.empty())
@@ -39,20 +41,13 @@ namespace ZividPython
                     auto previous = i - 1;
                     auto next = i + 1;
 
-                    // if surrounded by capital case (looking at the B character): ABC -> abc
-                    if(isupper(upperCamelCase[previous])
-                       && ((next < upperCamelCase.size() && isupper(upperCamelCase[next]))
-                           || (next >= upperCamelCase.size())))
+                    if(!isupper(upperCamelCase[previous])
+                       || (next < upperCamelCase.size() && !isupper(upperCamelCase[next])))
                     {
-                        ss << char(tolower(upperCamelCase[i]));
-                    }
-                    // all other cases results in adding an underscore first
-                    else
-                    {
-                        ss << "_" << char(tolower(upperCamelCase[i]));
+                        ss << "_";
                     }
+                    ss << char(tolower(upperCamelCase[i]));
                 }
-                // already lower case stays lower case: a -> a
                 else
                 {
                     ss << char(upperCamelCase[i]);
```

eskaur:
This is done now, right @anordal ?

anordal:
Yes, I see it's in master even. Closing.

See [StdExt.cpp](https://github.com/zivid/zivid-sdk/blob/master/cpp/Library/Base/StdExt.cpp#L59) for how it ended up in SDK.

PR #92, opened 2020-09-23
Minor 2.0 fixes

trym-b:


eskaur:
By the way, I just merged my Calibration fixes to the release-candidate-2.0 branch. Can you rebase? :)

anordal (approved):

eskaur:
Since we have no unit-tests for toSnakeCase: Has this been properly tested?

trym-b:
I haven't really tested this, I got some patch files from @anordal. Assuming that all tests passes, then I would think all of the functions/modules has the correct names.

eskaur:
Ah it is in fact used during build. For some reason I thought it was (almost) dead code since we haven't deployed the code-generation yet.

Never mind then. :)

anordal:
The patch is #91. Just an equivalence transform from my side. I haven't tested it by itself either.

eskaur (approved):

PR #94, opened 2020-10-02
Decrease precision requirement of hand-eye test

eskaur:
The reference matrix was based on a Linux run, and the outcome of this
algorithm seems to be a bit sensitive to the compiler/platform (since
the test-dataset has quite few poses). Changing the tolerance to let
this pass test on Windows.


eskaur:
Note that this is not a regression. This test has never passed on Windows before.

trym-b (approved):

PR #96, opened 2020-10-05
Fix linting

eskaur:


.flake8-packaged-files
```diff
@@ -10,6 +10,12 @@ ignore =
...
-    D402, # First line should not be the function’s “signature”
 show-source = True
+
+### Ignore docstring complaints in data models
+per-file-ignores =
+    settings.py:D101,D102,D106,D107
+    settings_2d.py:D101,D102,D106,D107
+    camera_state.py:D101,D102,D106,D107
+    camera_info.py:D101,D102,D106,D107
+    frame_info.py:D101,D102,D106,D107
```
trym-b, line 21: You probably want to add a comment about what kind of errors D101, D102, D106 and D107 really are. Makes it easier to fix them in the right order later

samples/sample_capture_assistant.py
```diff
@@ -10,7 +10,7 @@ def _main():
 
     suggest_settings_parameters = SuggestSettingsParameters(
         max_capture_time=datetime.timedelta(milliseconds=1200),
-        ambient_light_frequency=SuggestSettingsParameters.AmbientLightFrequency.none,
+        ambient_light_frequency="none",
```
trym-b, line 13: Why was this changed?

trym-b (commented):
Added some minor comments, looks good

samples/sample_capture_assistant.py
eskaur: See the relevant commit message:
```
I believe this is better, as it makes using the CA seem easier.  Strings
are the common way of inputting such arguments in Python anyway.
```

I don't feel strongly about this though, I have just found myself being annoyed at typing out a lot of stuff lately when wanting to use the CA in python, and just realized that they are strings now. 
Feel free to disagree, or say that it doesn't belong in this PR :)

.flake8-packaged-files
eskaur: Good idea, I'll do that.

eskaur: Done

samples/sample_capture_assistant.py
trym-b: It might be easier to discover the other values ambient light frequenct could be if there with the previous solution. If a user just wants to play around with this sample and wants to change the frequency to something else, the user`s IDE could potentially help them out.

.flake8-packaged-files
trym-b: Nice

samples/sample_capture_assistant.py
eskaur: Good point. I can revert this.

eskaur: Done.

eskaur:
Ready for next round, @trym-b

trym-b (approved):
Looks good to me, I am happy that this branch now passes the CI

eskaur:
Oh no, the ArchLinux build is set to always use the latest packages on AUR. I just upgraded AUR to 2.1 now, so any build targeting 2.0 will fail.

PR #99, opened 2020-10-08
Update README text

eskaur:


eskaur:
Kinda makes me wish that `camera.capture()` with no arguments just used default Settings...

trym-b (approved):

PR #102, opened 2020-10-27
Fix windows with python3.8

eskaur:


continuous-integration/windows/build.py
```diff
@@ -26,7 +26,7 @@ def _run_process(args):
 
 
 def _build(root):
-    _run_process(("pip", "install", str(root)))
+    _run_process(("pip", "install", "--verbose", str(root)))
```
trym-b, line 29: This looks like a debug print, delete?

trym-b (commented):
Looks good.

But what is `DLL hijacking`?

continuous-integration/windows/build.py
eskaur: Actually it was on purpose, because I wanted more output in the CI. If we had that, we would have caught the CI bug much sooner.

Do you think it should be removed?

eskaur:
> Looks good.
> 
> But what is `DLL hijacking`?

"DLL hijacking" would be if someone put a malicious "_zivid.dll" or something on the user's system, and zivid-python loaded that one instead. I don't think it's our responsibility to avoid that.

continuous-integration/windows/build.py
trym-b: If it is useful, then keep it. Just add a line about it in the commit message

eskaur: Alright

eskaur: Done. Ready for next round.

trym-b (approved):

Issue #4, opened 2019-07-02
Documentation for camera argument to firmware.update is wrong


nedrebo:
**Actual:** https://github.com/zivid/zivid-python/blob/1a4039a173bf54dc891a18e538df36e5c943add1/modules/zivid/firmware.py#L14

**Expected:** "camera: The camera to be updated."

eskaur:
Looks like this issue has been buried for a long while. Finally made a PR to resolve it.

PR #105, opened 2020-11-03
Issue #4: Fix firmware docstrings

eskaur:
The "Args" entries in the docstrings of update() and is_up_to_date()
were misleading. This commit fixes that.

Resolves Issue #4

### DIFF
--- a/modules/zivid/firmware.py
+++ b/modules/zivid/firmware.py
@@ -11,7 +11,7 @@ def update(camera, progress_callback=None):
     Call is_up_to_date() first to check if the camera is up to date.
 
     Args:
-        camera: A camera to check if firmware is up to date.
+        camera: The camera to be updated.
         progress_callback: A callback function to track progress of update.
             The callable is taking a float and a string as progress and description respectively.
     """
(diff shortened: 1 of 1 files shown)


trym-b (approved):

modules/zivid/firmware.py
```diff
@@ -11,7 +11,7 @@ def update(camera, progress_callback=None):
     Call is_up_to_date() first to check if the camera is up to date.
 
     Args:
-        camera: A camera to check if firmware is up to date.
+        camera: The Camera to be updated.
```
nedrebo, line 14: I do not think upper case is right here (or in the change further down in the file).

If you refer to the class Camera then you must write `Camera instance`, if this is a physical camera then lower case is correct.

If this is referring to the class I do not know if there are any docstring conventions that should be used so that generated code can produce the correct link.

nedrebo (changes requested):

modules/zivid/firmware.py
eskaur: You're right, it makes more sense to refer to it as a physical camera in this case, so I changed it to lower-case "camera".

We should look into more docstring conventions when we actually start generating docs, but I don't think that's needed at this point.

nedrebo (approved):

PR #109, opened 2020-12-19
Migrate CI from Azure Pipelines to Github Actions

eskaur:
This commit migrates the build+test+lint CI workflow from Azure
Pipelines to Github Actions. This is done in preparation for expanding
to more advanced workflows for e.g. deployment, since we know Github
Actions better as of now.

This commit also adds unit-testing on Archlinux, so that we may get
early warning if we ever stop working with the latest Python.


continuous-integration/windows/build.py
```diff
@@ -30,9 +21,8 @@ def _build(root):
 
 
 def _main():
-    options = _options()
-
-    _build(options.root)
+    root = Path(__file__).resolve().parents[2]
```
trym-b, line 24: I see we do this twice in two different scripts, could we somehow centralize this functionality?

continuous-integration/windows/test.py
```diff
@@ -59,10 +52,9 @@ def _install_pip_dependencies():
 
 
 def _main():
-    options = _options()
-
     _install_pip_dependencies()
-    _test(options.root)
+    root = Path(__file__).resolve().parents[2]
```
trym-b, line 56: Here is the other instance

trym-b (commented):
Nice initivative

README.md
```diff
@@ -151,12 +151,10 @@ Please visit [Zivid Knowledge Base](http://help.zivid.com) for general informati
...
 | Windows 10       | 3.6, 3.7, 3.8, 3.9 | 2.2.0             |
-| Arch Linux*      | latest             | 2.2.0             |
-
-[*] Only build, no unit testing.
+| Arch Linux       | latest             | 2.2.0             |
 
 [header-image]: https://www.zivid.com/hubfs/softwarefiles/images/zivid-one-plus-3x-1200x400.png
-[ci-badge]: https://img.shields.io/azure-devops/build/zivid-devops/376f5fda-ba80-4d6c-aaaa-cbcd5e0ad6c0/2/master.svg
-[ci-url]: https://dev.azure.com/zivid-devops/zivid-python/_build/latest?definitionId=2&branchName=master
+[ci-badge]: https://github.com/zivid/zivid-python/workflows/Main%20CI%20workflow/badge.svg?branch=master
```
nedrebo, line 157: https://github.com/zivid/zivid-repository-guidelines#build-badges

continuous-integration/windows/setup.py
```diff
@@ -58,9 +59,10 @@ def _install_intel_opencl_runtime():
 
 def _write_zivid_cpu_configuration_file():
     api_config = Path() / "ZividAPIConfig.yml"
-    target_location = Path(
-        r"C:\Users\VssAdministrator\AppData\Local\Zivid\API\Config.yml"
-    )
+
+    appdata_dir = Path(os.getenv("LOCALAPPDATA"))
+    target_location = appdata_dir / "Zivid" / "API" / "Config.yml"
+    print("Target location for config: " + str(target_location))
```
nedrebo, line 65: :+1:

continuous-integration/windows/build.py
eskaur: Not sure if it's worth the added complexity. It's a very simple one-liner, and we repeat this kind of thing all the time in the equivalent bash scripts.

README.md
eskaur: There's always some rule I don't know about ;)
How about now? (pushed a fixup) You can change the branch-name from master to this branch to test.

nedrebo (dismissed):

README.md
nedrebo: Beautiful 👌

azure-pipelines.yml
```diff
@@ -1,69 +0,0 @@
-schedules:
-- cron: "0 22 * * *"
-  displayName: Master nightly build
-  branches:
-    include:
-    - master
-  always: true
```
nedrebo, line 7: This functionality is lost now. We should keep it as it is a nice mechanism to detect when for example a new Python breaks on Arch.

eskaur: Ah, i'll see if I can get it re-introduced. I assume one can trigger on both push and timer.

eskaur: Fixed now.

nedrebo (approved):

continuous-integration/windows/build.py
trym-b: Ok, we can probably refactor it later if we need to

trym-b (approved):

eskaur:
- Squashed.
- Changed the "required checks" of master branch to the ones introduced by this PR.

trym-b (approved):

PR #112, opened 2021-01-08
Add Github Actions workflow to assist release deployment

eskaur:
This commit adds a new workflow called deploy.yml which creates
deployable artifacts for PyPI (or direct installation with pip). The
workflow is triggered on push to master branch (which includes any time
a PR towards master is merged). This choice was made because it is
convenient that up-to-date artifacts always are available, even if there
is no impending release.

The workflow does not go as far as actually upload to PyPI. The purpose
of the workflow is to provide, as an artifact, a zip-file which contains
the full set of tested and validated wheels/source-dists ready for
upload to PyPI. A developer with the necessary credentials may then
simply download the artifact called "dist_all" and run `twine upload *`
on the unzipped contents.

Workflow map:
image

Resulting persistent artifact:
image


nedrebo:
It is a good idea to make all the artifacts that are going to be deployed to PYPI, even if the deployment is not automated yet.

However, I do not understand why workflow is now split in two. All the testing that is done in the original workflow on the public interface should now instead be tested using the artifacts. That would be a more realistic and contained test, now we are testing on binaries that are not going to be shipped.

As long as this is actually not deployment I would merge the workflows and move the building of artifacts to the beginning of the flow and remove the additional (existing) build and instead just do a pip install of one of the artifacts.

Also since this is not actual deployment I think we can and should do everything on all pushes.

eskaur:
> It is a good idea to make all the artifacts that are going to be deployed to PYPI, even if the deployment is not automated yet.
> 
> However, I do not understand why workflow is now split in two. All the testing that is done in the original workflow on the public interface should now instead be tested using the artifacts. That would be a more realistic and contained test, now we are testing on binaries that are not going to be shipped.
> 
> As long as this is actually not deployment I would merge the workflows and move the building of artifacts to the beginning of the flow and remove the additional (existing) build and instead just do a pip install of one of the artifacts.
> 
> Also since this is not actual deployment I think we can and should do everything on all pushes.

Yeah you're right. It started as a separate workflow because I was aiming for PyPI deployment, but in its current state it does make more sense to merge the two workflows. I got blinded by the original design :)

eskaur:
@nedrebo  Something like this?:
image

The new main.yml is basically my old deploy.yml + Linting.

nedrebo:
That image looks good. However, I did not look into de details to see if we lost a "class" of tests in the process. I think @trym-b has a better overview of the current test coverage.

From afar this looks :+1:

.github/workflows/main.yml
```diff
@@ -7,7 +7,75 @@ on:
     - cron:  '0 22 * * *'
 
 jobs:
-  Linux-build-and-test:
+  Lint:
```
trym-b, line 10: Since we are changing the names of these jobs, could we transform all of these names in to lower case?

.github/workflows/main.yml
```diff
@@ -7,7 +7,75 @@ on:
...
+              --workdir /host/continuous-integration/linux \
+              --env "PYTHONDONTWRITEBYTECODE=1" \
+              ubuntu:20.04 \
+              bash -c "./setup.sh && ./create-sdist.sh"
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: sdist
+          path: dist/zivid*.tar.gz
+          retention-days: 1
```
trym-b, line 46: Is it possible to specify hours instead of days? 1 hour should be enough

.github/workflows/main.yml
```diff
@@ -7,7 +7,75 @@ on:
...
+      - name: Build and lint on Archlinux
+        run: | 
+          docker run \
+              --volume $PWD:/host \
+              --workdir /host/continuous-integration/linux \
+              --env "PYTHONDONTWRITEBYTECODE=1" \
+              archlinux/base \
+              bash -c "./setup.sh && ./build.sh && ./lint.sh"
+
+  Create-source-dist:
```
trym-b, line 26: How about `create-source-distribution`?

continuous-integration/deployment/check_expected_artifacts.py
```diff
@@ -0,0 +1,30 @@
+from pathlib import Path
+
+
+def _get_expected_artifacts():
```
trym-b, line 4: I think we can drop the `_get` prefix here since we are just reading a file plus some more

continuous-integration/deployment/check_expected_artifacts.py
```diff
@@ -0,0 +1,30 @@
...
+    artifacts_file = Path(__file__).resolve().parent / "expected-artifacts.txt"
+    artifacts = artifacts_file.read_text().splitlines()
+    artifacts.sort()
+    version_file = Path(__file__).resolve().parent / "expected-version.txt"
```
trym-b, line 8:
```suggestion
    parent_dir = Path(__file__).resolve().parent
    version_file = parent_dir / "expected-version.txt"
```

continuous-integration/linux/collect-and-check-artifacts.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+CI_DIR=$(realpath "$SCRIPT_DIR/..")
+ROOT_DIR=$(realpath "$SCRIPT_DIR/../..")
+
+# Simple setup
+apt-get update && apt-get install --yes python3 python3-pip || exit $?
```
trym-b, line 8: This can be split into two lines

continuous-integration/linux/collect-and-check-artifacts.sh
```diff
@@ -0,0 +1,19 @@
...
+pip3 install twine || exit $?
+
+# Move artifacts into a common directory
+mkdir $ROOT_DIR/dist || exit $?
```
trym-b, line 12: Always quote all paths in case there is a whitespace present

continuous-integration/linux/create-sdist.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
trym-b, line 1: Could you rename this file to something with less shorthands? `create-source-distribution.sh` perhaps?

continuous-integration/linux/create-sdist.sh
```diff
@@ -0,0 +1,12 @@
...
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT_DIR=$(realpath "$SCRIPT_DIR/../..")
+
+# Install minimal requirements to create the source distributions
+pip install --requirement "$SCRIPT_DIR/../python-requirements/build.txt" || exit $?
+
+# Create source distribution
+(cd $ROOT_DIR && python setup.py sdist) || exit $?
```
trym-b, line 10: Can this line be split as well?

continuous-integration/linux/install-sdist.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
trym-b, line 1: Same with this filename, could we use longer names? It might make it easier for newer people to contribute

continuous-integration/linux/install-sdist.sh
```diff
@@ -0,0 +1,10 @@
...
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT_DIR=$(realpath "$SCRIPT_DIR/../..")
+
+# Install source distribution
+pip install --upgrade pip
```
trym-b, line 7: Missing: `|| exit $?`

trym-b: Could you also run `python(3?) -m pip install --upgrade pip`?

continuous-integration/linux/install-sdist.sh
```diff
@@ -0,0 +1,10 @@
...
+pip install $ROOT_DIR/dist/*.tar.gz || exit $?
```
trym-b, line 8: Remember to quote all path arguments in case there is a whitespace in it

continuous-integration/windows/create_bdist.py
```diff
@@ -0,0 +1,13 @@
+from common import repo_root, run_process, install_pip_dependencies
```
trym-b, line 1: Could we also rename this file?

continuous-integration/windows/install_bdist.py
```diff
@@ -0,0 +1,34 @@
+from common import repo_root, run_process
+
+
+def _find_wheel(root):
+    extension = ".whl"
+    dist_dir = root / "dist"
+    wheels = list(dist_dir.glob("*" + extension))
```
trym-b, line 7: So sad we cannot use f-strings here...

trym-b (commented):
Looks good to me

.github/workflows/main.yml
eskaur: Sure.

.github/workflows/main.yml
eskaur: I was looking for something like that, but I could only find the argument "retention-days", no "retention-hours". 
I think 1 day should be fine though, the artifacts are like 1MB.

.github/workflows/main.yml
eskaur: Sure. I'll change all the other "dist" to "distribution" too then.

continuous-integration/deployment/check_expected_artifacts.py
eskaur: Sure. Same for the other function too then.

continuous-integration/linux/collect-and-check-artifacts.sh
eskaur: Will do.

continuous-integration/linux/collect-and-check-artifacts.sh
eskaur: I'll try to do that where I've added paths. It gets a bit awkward in some places because asterisks (*) are not expanded, and I have to put them outside of the quotes.

continuous-integration/linux/create-sdist.sh
eskaur: Sure. Same for the others also then.

continuous-integration/linux/install-sdist.sh
eskaur: Renaming to "build-and-install-source-distribution.sh", since it actually does a complete build based on the sdist.

continuous-integration/linux/create-sdist.sh
eskaur: I'd like to avoid that. The only reason why I do "cd" is that setup.py must be called from the same directory as the file itself, and I don't need the directory change to live beyond this line. The benefit of doing it this way is that the directory change is done in a subshell, so the working directory is unchanged as the script moves on.

continuous-integration/linux/install-sdist.sh
eskaur: Yeah that was missing, thanks.

Regarding pip: I followed the convention from build.sh, where we use straight forward pip commands. These agents only have one pip installed. Do you have a good argument for complicating the command?

continuous-integration/linux/install-sdist.sh
eskaur: Will do. I have to end the quote before the asterisk (*) though. If not it doesn't work (doesn't get expanded).

continuous-integration/windows/create_bdist.py
eskaur: Renamed these files to "create_binary_distribution.py" and "install_binary_distribution.py".

continuous-integration/windows/install_bdist.py
eskaur: Yup.. :)

eskaur:
Ready for next round.

nedrebo (dismissed):
I approve of the overall re-design :+1:

However, I have not reviewed the details. For these kinds of refactoring PRs it is especially important to check that we did not lose any test coverage in the process. I'll leave approving that to @trym-b.

.github/workflows/main.yml
trym-b: Ok, sounds good

continuous-integration/linux/create-sdist.sh
trym-b: Perhaps a comment might be needed? I did not think of this when I read this part of the code.

continuous-integration/linux/install-sdist.sh
trym-b: This [link](https://snarky.ca/why-you-should-use-python-m-pip/) explains it nicely

.github/workflows/main.yml
```diff
@@ -17,16 +85,22 @@ jobs:
         uses: actions/checkout@v2
         with:
           lfs: true
-      - name: Build and test on ${{matrix.os}}
+      - name: Download artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: sdist
```
trym-b, line 91: Nitpick: rename the artifacts we upload to `source-distribution`?

.github/workflows/main.yml
```diff
@@ -17,16 +85,22 @@ jobs:
...
+          path: dist
+      - name: Install from sdist and test
```
trym-b, line 93:
```suggestion
      - name: Install from source distribution and test
```

.github/workflows/main.yml
```diff
@@ -36,33 +110,44 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          name: bdist-win-python${{matrix.python-version}}
+          path: dist
       - name: Set up Python ${{matrix.python-version}}
         uses: actions/setup-python@v2
         with:
           python-version: ${{matrix.python-version}}
-      - name: Setup
+      - name: Setup
```
trym-b, line 122: Super nitpick:
```suggestion
      - name: Setup
```
We should consider adding some yaml file formatting

continuous-integration/deployment/check_expected_artifacts.py
```diff
@@ -0,0 +1,31 @@
...
+    artifacts_file = parent_dir / "expected-artifacts.txt"
+    artifacts = artifacts_file.read_text().splitlines()
+    artifacts.sort()
+    version_file = parent_dir / "expected-version.txt"
+    version = version_file.read_text().strip()
+    return [artifact.format(version=version) for artifact in artifacts]
+
+
+def _present_artifacts():
+    artifacts_dir = Path(__file__).resolve().parents[2] / "dist"
```
trym-b, line 15:
```suggestion
    artifacts_dir = Path(__file__).resolve().parents[2] / "distribution"
```
The reason for this is that it makes the workflow cleaner, the path itself becomes `distribution`

trym-b (commented):
Added some more naming nitpick comments. Looks very clean in general

continuous-integration/linux/create-sdist.sh
eskaur: Sure I can add a comment.

continuous-integration/linux/install-sdist.sh
eskaur: Alright I'll change it across the board then.

.github/workflows/main.yml
eskaur: Will do.

.github/workflows/main.yml
eskaur: Will do

.github/workflows/main.yml
eskaur: Will do

continuous-integration/deployment/check_expected_artifacts.py
eskaur: Sure, done.

eskaur:
Ready for next round @trym-b .

eskaur:
Squashed and rewrote the commit message, since the original message was not applicable to the latest design.

trym-b (approved):
Nice work.

nedrebo (approved):

PR #113, opened 2021-01-12
MISC: Use generic header image

SatjaSivcev:


eskaur (approved):
Thanks for dealing with this.

PR #114, opened 2021-01-28
Bump Ubuntu 16.04 to Python 3.6

eskaur:
The default python3 of Ubuntu 16.04 (python3.5) has been end-of-life
since September 2020. Recently there has been some upgrades to
dependencies on PyPI that causes zivid-python to stop working with
U16/python3.5. This includes numpy, as well as pip itself (it breaks
itself when upgrading if on python3.5).

While it may be possible to work around this to retain U16/python3.5
support, it seems more reasonable to drop python3.5 support altogether
at this point. This commit adjusts ubuntu-16.04/setup.sh so that
python3.6 is used instead.

It was also discovered that the installation of lsb-core now resets g++
to point to g++-5, so the update-alternatives lines were moved further
down in ubuntu-16.04/setup.sh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Zivid Python is the official Python package for Zivid 3D cameras. Read more abou
 
 ### Dependencies
 
-* [Python](https://www.python.org/) version 3.5 or higher
+* [Python](https://www.python.org/) version 3.6 or higher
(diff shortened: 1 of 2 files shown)


eskaur:
We have decided to just drop U16 support from the CI instead.

PR #119, opened 2021-02-02
CI: Enable CI for external pull requests

trym-b:
Previously, if someone forked this repo and created a pull request from
that external repo, Github Actions would not be triggered. This makes it
difficult to know if the external changes will break this repo or not.

This commit ensures that external PRs now will trigger CI builds.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,8 +2,9 @@ name: Main CI workflow
 # Note: If the workflow name is changed, the CI badge URL in the README must also be updated
 
 on:
-  push:       # Push trigger runs on any pushed branch.
-  schedule:   # Scheduled trigger runs on the latest commit on the default branch.
+  push:         # Push trigger runs on any pushed branch.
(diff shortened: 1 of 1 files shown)


eskaur:
Hm, why is the "checks" list listing everything double?..

trym-b:
I think it is because the same commit is triggered twice, once by the push and once by creating the PR.

eskaur:
> I think it is because the same commit is triggered twice, once by the push and once by creating the PR.

Hm, that's not ideal, and also not how I thought it would work. Should look into that before merge.

trym-b:
It seems like it should be okay according to the documentation: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-using-a-list-of-events

eskaur (approved):
I've been assured that the "double-CI" is a temporary quirk of this PR, and that it will be gone once the change is on master.

If that turns out to be untrue, we just revert this.

trym-b:
Yeah, let's try it.

PR #121, opened 2021-02-17
CI: Use runc v1.0.0-rc93 on hosts with archlinux containers

knzivid:
Can be reverted after these issues are fixed.

https://github.com/actions/virtual-environments/issues/2658
https://github.com/moby/moby/pull/42014

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,13 +14,20 @@ jobs:
         uses: actions/checkout@v2
         with:
           lfs: false
+      # TODO(#122): Remove step when issue #122 is solved
+      - name: update runc # https://github.com/actions/virtual-environments/issues/2698#issuecomment-779262068
+        run: |
+          sudo apt-get install --assume-yes libseccomp-dev
+          git clone https://github.com/opencontainers/runc
+          cd runc && git checkout v1.0.0-rc93 && make -j`nproc` && sudo make install
+          cd .. && rm --recursive runc
       - name: Build and lint on Archlinux
-        run: | 
+        run: |
           docker run \
               --volume $PWD:/host \
               --workdir /host/continuous-integration/linux \
               --env "PYTHONDONTWRITEBYTECODE=1" \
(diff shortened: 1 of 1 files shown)


eskaur:
Excellent. I can fix the "required tests" list so it's possible to merge.

knzivid:
It's already green

.github/workflows/main.yml
```diff
@@ -14,13 +14,19 @@ jobs:
         uses: actions/checkout@v2
         with:
           lfs: false
+      - name: build latest runc # https://github.com/actions/virtual-environments/issues/2698
```
eskaur, line 17: You are referring to a different issue here than in the commit message. Was that intentional?

eskaur:
> It's already green

Yes, because I removed "archlinux/base" as required check. It was stuck on Yellow.

.github/workflows/main.yml
eskaur, line 17: Can you explicitly write "Remove step when {...} is solved" on both instances of "build latest runc" here?

knzivid: The issues in the commit message point to this issue's comment as a workaround.

knzivid: I created #122 to follow up on this PR. I can add a TODO(#122) in both places

eskaur: Ah yes, I should have looked more closely. :)

knzivid: I updated the yaml comment as well as the commit message to say this explicitly. Hope its better now :)

eskaur:
```
# TODO(#122): Remove step when PR #121 is solved
```

Don't you mean `# TODO(#122): Remove step when Issue #122 is solved`

knzivid: oh yeah, will fix

eskaur (approved):
Thanks for dealing with this. I'll remember to follow up on #122.

PR #123, opened 2021-02-26
Fix bug where __bool__ -> True when valid() -> False

eskaur:
The classes in the calibration API have the following implementation on
the front-end,
```
def __bool__(self):
    return bool(self.__impl)
```
which was intended to represent the corresponding boolean operator on
the C++ API level. However, the boolean operator of the `__impl`s were
never overridden on the Pybind11 level, so they instead defaulted to
some behavior that seemingly always returns True, and is independent of
the `operator bool` in the C++ API. This caused the following kind of
bugs in the Python API
```
detection_result.valid() # False
bool(detection_result)   # True
```
This commit fixes this by explicitly hooking up the `__bool__` of the
Pybind11-generated classes to the `operator bool` of the C++ classes.
This yields behavior that is identical to the C++ API, as intended.

No unit-tests are added for this because instantiating invalid
calibration objects like DetectionResult requires running
detect_feature_points with a non-checkerboard point-cloud, and this
would unreasonably slow down the CI.

### DIFF
--- a/src/include/ZividPython/Traits.h
+++ b/src/include/ZividPython/Traits.h
@@ -1,0 +1,45 @@
+#pragma once
+
+#include <type_traits>
+
+#if __has_include(<experimental/type_traits>)
+#    include <experimental/type_traits>
+#else
+namespace Detail
+{
+    // Fallback implementation of is_detected for compilers lacking <experimental/type_traits>
+    struct nonesuch
+    {
+        ~nonesuch() = delete;
+        nonesuch(nonesuch const &) = delete;
+        void operator=(nonesuch const &) = delete;
+    };
+
+    template<class Default, class AlwaysVoid, template<class...> class Op, class... Args>
+    struct detector
+    {
+        using value_t = std::false_type;
+        using type = Default;
+    };
+
+    template<class Default, template<class...> class Op, class... Args>
+    struct detector<Default, std::void_t<Op<Args...>>, Op, Args...>
+    {
+        using value_t = std::true_type;
+        using type = Op<Args...>;
+    };
+} // namespace Detail
+#endif
+
+namespace ZividPython
+{
+#if __has_include(<experimental/type_traits>)
+    // Has the experimental header, so just use that.
+    template<template<class...> class Op, class... Args>
+    using is_detected = typename std::experimental::is_detected<Op, Args...>;
+#else
+    // Use fallback implementation.
+    template<template<class...> class Op, class... Args>
+    using is_detected = typename Detail::detector<Detail::nonesuch, void, Op, Args...>::value_t;
+#endif
+} // namespace ZividPython
--- a/src/include/ZividPython/Wrappers.h
+++ b/src/include/ZividPython/Wrappers.h
@@ -10,6 +10,7 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include <ZividPython/Traits.h>
 #include <ZividPython/Wrapper.h>
 
 #include <pybind11/pybind11.h>
@@ -18,6 +19,9 @@ namespace ZividPython
 {
     namespace
     {
+        template<typename T>
+        using bool_t = decltype(static_cast<bool>(std::declval<T>()));
+
         /// Inserts underscore on positive case flank and before negative case flank:
         /// ZividSDKVersion -> zivid_sdk_version
         std::string toSnakeCase(const std::string upperCamelCase)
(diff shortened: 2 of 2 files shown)


src/Calibration/Detector.cpp
```diff
@@ -12,6 +12,7 @@ namespace ZividPython
     void wrapClass(pybind11::class_<Zivid::Calibration::DetectionResult> pyClass)
     {
         pyClass.def("valid", &Zivid::Calibration::DetectionResult::valid)
+            .def("__bool__", &Zivid::Calibration::DetectionResult::operator bool)
```
nedrebo, line 15: This is fine, but even better is to use `constexper` and `std::is_detected` and add `"__bool__"` to all wrapped classes automatically when supported.

You can do that in https://github.com/zivid/zivid-python/blob/master/src/include/ZividPython/Wrappers.h#L75 and then this is fixed once and for all.

nedrebo (commented):
You should add a unit test to avoid regressions.

eskaur:
> You should add a unit test to avoid regressions.

As mentioned in the commit message, that would cost quite a bit of CI time.
I added a failed-detection unit-test way back, but had to remove it because you thought the CI took too long. Are you changing your mind on that?

src/Calibration/Detector.cpp
eskaur: Good idea. I'll give it a shot.

nedrebo:
> > You should add a unit test to avoid regressions.
> 
> As mentioned in the commit message, that would cost quite a bit of CI time.
> I added a failed-detection unit-test way back, but had to remove it because you thought the CI took too long. Are you changing your mind on that?

Ok, sry I missed that.

src/Calibration/Detector.cpp
eskaur: I just pushes a fixup that does this.

Problem: The compiler on Windows doesn't have `<experimental/type_traits>`.
Possible solution: Just copy in the "possible implementation" from https://en.cppreference.com/w/cpp/experimental/is_detected ? 

What do you think?

nedrebo: I think that is fine. I would add it to a separate header and use `#if __has_include` so the unofficial implementation is only used when needed.

eskaur: Added a new solution now. Please have a close-ish look, because I'm on super thin ice here, C++-wise...

Also temporarily added a unit test that used to fail and now succeeds. It adds about 2 minutes to the CI (+14%).

src/include/ZividPython/Traits.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include <type_traits>
+
+#if __has_include(<experimental/type_traits>)
+#    include <experimental/type_traits>
+#endif
```
nedrebo, line 7:
```suggestion
#else
```

Hide the fallback when not needed.

src/include/ZividPython/Traits.h
```diff
@@ -0,0 +1,45 @@
...
+
+namespace
```
nedrebo, line 9:
```suggestion
namespace Detail
```

https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file

nedrebo: There is a clang-tidy check for this, consider enabling it, if it is not already enabled.

nedrebo: https://clang.llvm.org/extra/clang-tidy/checks/google-build-namespaces.html

src/include/ZividPython/Traits.h
eskaur: Ah you mean pack the entire implementation (Detail namespace) in the else block?

src/include/ZividPython/Traits.h
eskaur: Ah yes, I knew that. Forgot I was in a header...

I realize we don't run clang-tidy at all in this repo yet. Enabling that seems to me as too much a scope-creep for a bugfix PR. Do you agree?

src/include/ZividPython/Traits.h
eskaur: Done

src/include/ZividPython/Traits.h
eskaur: Anyway, put it in a Detail namespace now.

nedrebo: We should enable it then, but that is a task in itself.

nedrebo (approved):

eskaur:
Stale review due to fixup-squash. Gonna need a re-approve, @nedrebo.

nedrebo (approved):

Issue #122, opened 2021-02-17
Undo PR #121 once actions/virtual-environments issue 2658 is fixed


knzivid:
Track https://github.com/actions/virtual-environments/issues/2658

eskaur:
Once this is all settled, would you recommend going back to "archlinux/base" or something else?

knzivid:
Nah.. I would still stick to archlinux/archlinux. The whole point of arch is to keep rolling and directly using the images from archlinux upstream is better

eskaur:
Aha, so will revert from `archlinux/archlinux:base-devel` to `archlinux/archlinux:base` when resolving this issue.

knzivid:
We were originally on the docker hub packaged `archlinux` but we switched now to `archlinux/archlinux`. The reason why we switched to `base-devel` tag is that we build AUR packages and

> base-devel is a package group that includes tools needed for building (compiling and linking). It is not necessary for a basic install, and many users don't need to install it. If you find it useful, you can either install it as part of your basic install, or later.

eskaur:
Resolved by https://github.com/zivid/zivid-python/pull/125

PR #125, opened 2021-03-07
Revert pr121 arch fix

eskaur:
Note how both commits in this PR are green (it now works with both `archlinux/base` and `archlinux/archlinux:base`), so the second commit is not strictly necessary. However, I added the second commit due to the discussion in #122 .
I'll let you Arch experts decide what's best.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,20 +14,13 @@ jobs:
         uses: actions/checkout@v2
         with:
           lfs: false
-      # TODO(#122): Remove step when issue #122 is solved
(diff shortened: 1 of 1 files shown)


knzivid:
Let's stick to arch's arch image

knzivid:
Like I previously mentioned, we are already rolling; why not roll with upstream?

eskaur:
> Like I previously mentioned, we are already rolling; why not roll with upstream?

Guess I need approval from a "code owner". @nedrebo ?

nedrebo (approved):

PR #126, opened 2021-04-14
Bump to SDK 2.3

eskaur:


setup.py
```diff
@@ -8,11 +8,11 @@
 
 # To be replaced by: from setuptools_scm import get_version
 def get_version():
-    return "2.1.0"
+    return "2.1.1"
```
nedrebo, line 11: There are no wrapper changes as far as I can see. You can leave this number unchanged.

eskaur: But as mentioned in the relevant commit message, there have been bug-fixes in zivid-python itself since last release. Isn't this what the Patch Version number is for?

nedrebo: Sorry I missed that, then nevermind :o

nedrebo (approved):

PR #129, opened 2021-04-30
Test

eskaur:
This is just a test-PR. Do not merge.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,4 @@
+"""This is a test-PR to ensure that we do not get duplicated CI-jobs due to the recent change."""
 import tempfile
 import platform
 import subprocess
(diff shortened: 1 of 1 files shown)


eskaur:
@trym-b : Got duplicate workflows when pushing an update to this branch, which contradicts the premise for merging https://github.com/zivid/zivid-python/pull/119.

trym-b:
That is unfortunate. And also surprising. Not sure how we are going to let forks build the CI for `zivid-python`.

eskaur:
What about instead allowing manual triggering with "workflow_dispatch" that we can do in the rare case of an external PR?

trym-b:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-using-a-list-of-events

Perhaps there are some subtle differences when we use a different syntax?

eskaur:
Does it say anywhere in the docs that we should not expect duplicate workflows when multiple conditions are met?

trym-b:
The other syntax did not work: https://github.com/zivid/zivid-python/pull/130

There must be a correct way to do this.

trym-b:
I give up, I will create a PR that revert the previous commit

trym-b:
https://github.com/zivid/zivid-python/pull/131

Issue #93, opened 2020-09-23
toSnakeCase: string_view ftw


anordal:
```diff
--- a/src/include/ZividPython/DataModelWrapper.h
+++ b/src/include/ZividPython/DataModelWrapper.h
@@ -154,8 +154,7 @@ namespace ZividPython
 
                     using MemberType = std::remove_const_t<std::remove_reference_t<decltype(member)>>;
 
-                    std::string name{ MemberType::name };
-                    name = toSnakeCase(name);
+                    const std::string name = toSnakeCase(MemberType::name);
 
                     pyClass.def_property(
                         name.c_str(),
--- a/src/include/ZividPython/Wrappers.h
+++ b/src/include/ZividPython/Wrappers.h
@@ -14,13 +14,15 @@
 
 #include <pybind11/pybind11.h>
 
+#include <string_view>
+
 namespace ZividPython
 {
     namespace
     {
         /// Inserts underscore on positive case flank and before negative case flank:
         /// ZividSDKVersion -> zivid_sdk_version
-        std::string toSnakeCase(const std::string upperCamelCase)
+        std::string toSnakeCase(std::string_view upperCamelCase)
         {
             if(upperCamelCase.empty())
             {
@@ -29,12 +31,14 @@ namespace ZividPython
 
             if(!isupper(upperCamelCase[0]))
             {
-                throw std::runtime_error{ "First character of string: '" + upperCamelCase + "' is not capitalized" };
+                std::stringstream msg;
+                msg << "First character of string: '" << upperCamelCase << "' is not capitalized";
+                throw std::runtime_error{ msg.str() };
             }
             std::stringstream ss;
             ss << char(tolower(upperCamelCase[0]));
 
-            for(auto i = 1; i < upperCamelCase.size(); ++i)
+            for(size_t i = 1; i < upperCamelCase.size(); ++i)
             {
                 if(isupper(upperCamelCase[i]))
                 {
@@ -96,8 +100,8 @@ namespace ZividPython
                                   const WrapFunction &wrapFunction,
                                   const char *nonLowercaseName)
     {
-        std::string name{ nonLowercaseName };
-        auto submodule = dest.def_submodule(toSnakeCase(name).c_str());
+        const std::string name = toSnakeCase(nonLowercaseName);
+        auto submodule = dest.def_submodule(name.c_str());
         wrapFunction(submodule);
     }
 } // namespace ZividPython
```

eskaur:
Should this issue be closed, @anordal ?

[...]

PR #132, opened 2021-06-14
DataModelWrapper::toSnakeCase: Take string_view

anordal:
This function does not modify its string argument,
i.e. can use std::string_view, assuming C++17.
This avoids implicit std::string temporaries.

Also fixes a use-after-free of the form:

    toSnakeCase(name).c_str()

This aligns the function with the one in SDK,
including a couple of unrelated changes:

* Loop variable type: int → size_t (would satisfy gcc -Wsign-compare)
* The call to tolower can be made unconditional due to idempotence.

Implements #93

### DIFF
--- a/src/include/ZividPython/DataModelWrapper.h
+++ b/src/include/ZividPython/DataModelWrapper.h
@@ -156,8 +156,7 @@ namespace ZividPython
 
                     using MemberType = std::remove_const_t<std::remove_reference_t<decltype(member)>>;
 
-                    std::string name{ MemberType::name };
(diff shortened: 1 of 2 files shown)


nedrebo (approved):

anordal:
(I split the commit in 3).

Sorry, Mathias, I see you had some review comment(s), but they went away.

nedrebo (approved):
I had zero comments (nothing is lost) :+1:

anordal:
Moved to #133: The CI jobs don't start when the PR is from an external repository.

PR #144, opened 2021-10-01
Bump `black` to lastest version

trym-b:


modules/zivid/camera_state.py
```diff
@@ -16,39 +16,69 @@ def __init__(
...
                     )
                 )
-            if isinstance(general, (float, int,)):
+            if isinstance(
+                general,
+                (
+                    float,
+                    int,
+                ),
+            ):
```
eskaur, line 39: Is this really what black thinks is nice? :)

trym-b: > Black used to remove the trailing comma if the expression fits in a single line, but this was changed by #826 and #1288. Now a trailing comma tells Black to always explode the expression. This change was made mostly for the cases where you _know_ a collection or whatever will grow in the future. Having it always exploded as one element per line reduces diff noise when adding elements. Before the "magic trailing comma" feature, you couldn't anticipate a collection's growth reliably since collections that fitted in one line were ruthlessly collapsed regardless of your intentions. One of Black's goals is reducing diff noise, so this was a good pragmatic change.

https://black.readthedocs.io/en/stable/contributing/issue_triage.html?highlight=comma#the-trailing-comma-isn-t-being-removed-by-black

eskaur (approved):

PR #145, opened 2021-10-04
Bump linting tools

eskaur:
This commit does the following:
- Bump linting tools with `pip-compile --upgrade`.
- Update code to conform with new rules.

We ignore the "consider-using-f-string" rule because we do not want to
intentionally break the unofficial support for Ubuntu 16.04 / python3.5
yet. (f-strings were introduced in python3.6)

### DIFF
--- a/.pylintrc-tests
+++ b/.pylintrc-tests
@@ -1,5 +1,10 @@
 [MESSAGES CONTROL]
-disable=missing-docstring,import-outside-toplevel,duplicate-code,unused-argument,pointless-statement
+disable=missing-docstring,
+        import-outside-toplevel,
+        duplicate-code,
+        unused-argument,
+        pointless-statement,
+        consider-using-f-string # Keep .format() as long as we want unofficial python3.5 support
 
 [FORMAT]
 max-line-length=120
--- a/modules/zivid/camera.py
+++ b/modules/zivid/camera.py
@@ -3,10 +3,10 @@ from zivid.frame import Frame
 from zivid.frame_2d import Frame2D
 import zivid
 from zivid.settings_2d import Settings2D
-import zivid._settings_converter as _settings_converter
-import zivid._settings2_d_converter as _settings_2d_converter
-import zivid._camera_state_converter as _camera_state_converter
-import zivid._camera_info_converter as _camera_info_converter
+from zivid import _settings_converter
+from zivid import _settings2_d_converter
+from zivid import _camera_state_converter
+from zivid import _camera_info_converter
 import _zivid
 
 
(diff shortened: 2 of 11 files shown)


eskaur:
Immediate reason for this PR: 
Hit on a pylint-bug during development of the datamodel generator. This bug was fixed in the latest version of pylint.

.pylintrc-tests
```diff
@@ -1,5 +1,10 @@
 [MESSAGES CONTROL]
-disable=missing-docstring,import-outside-toplevel,duplicate-code,unused-argument,pointless-statement
+disable=missing-docstring,
+        import-outside-toplevel,
+        duplicate-code,
+        unused-argument,
+        pointless-statement,
+        consider-using-f-string # Keep .format() as long as we want legacy python3.5 support
```
nedrebo, line 7: Minor: I think you should drop the `legacy` word, I feel it adds nothing except a bit of a negative tone.

modules/zivid/camera.py
```diff
@@ -62,7 +62,7 @@ def capture(self, settings):
         if isinstance(settings, Settings2D):
             return Frame2D(
                 self.__impl.capture(
-                    _settings_2d_converter.to_internal_settings2_d(settings)
+                    _settings2_d_converter.to_internal_settings2_d(settings)
```
nedrebo, line 65: What forced you to do this? I hate it :O In order of preference: `_settings_2d` >> `_settings_2_d` >>>>> `_settings2_d`.

.pylintrc-tests
eskaur: I agree, it's a bit negative.
I felt like I couldn't just say `as long as we want python3.5 support` since we don't officially support python3.5 anymore. I wanted some modifier to reflect that. What about `unofficial python3.5 support`?

modules/zivid/camera.py
eskaur: Haha yes, this name is the worst. This function name has been like this since the release of zivid-python 2.0, when Settings2D was created partly by a code-generator. It comes from this:

```
>>> inflection.underscore("Settings2D")
'settings2_d'
```
Note that I did not mess with the module in this PR. I merely changed the import line at the top of the file that was used to locally rename the module.
We have an opportunity to improve these names in my upcoming code-generation PR, since these modules are private, but it seems out of the scope of this PR.

.pylintrc-tests
nedrebo: How about `# Keep .format() as long as we support python3.5`?

modules/zivid/camera.py
nedrebo: I agree, not in scope.

nedrebo (dismissed):

eskaur:
Need re-approval due to squash+forcepush :\ 
@nedrebo

nedrebo (approved):

PR #146, opened 2021-10-06
Automatically generate front-end modules for datamodels

eskaur:
While the Pybind11 equivalents of Zivid C++ data-models are
automatically generated in the "_zivid" back-end module, it has been
necessary to manually write/update the front-ends in the public "zivid"
module. This means that unnecessary manual work has been needed every
time the C++ data-models have been updated, and as a consequence of that
not always being prioritized this has lead to lack of support for the
most recent additions.

This improvement automatically adds support for the following settings
that already existed in the back-end, but had not been added to the
front-end:
- zivid.Settings.Experimental.Engine (enum type)
- zivid.Settings.Processing.Color.Experimental.ToneMapping (enum type)

This commit adds the following scripts:

`datamodel_frontend_generator.py --dest-dir <destination>`:
Core implementation that generates source code for a hard-coded list of data
models into a target directory.

`generate_datamodels.sh`:
Script for installing dependencies and generating code in `modules/zivid/`

`docker_generate_datamodels.sh`:
Script for running `generate_datamodels.sh` in Docker, to enable consistent results
on any platform.

`check_datamodels_up_to_date.py`:
Script to check that the source code in `modules/zivid` is up to date with the code
that can be generated by `datamodel_frontend_generator.py`.

There are two reasons why the modules are committed and not generated as
part of the build step:
- The only documentation supplied to customers is "look at the source
code in modules/zivid/".  Given this, it is better that the data models
can be readily inspected on Github.
- The script `datamodel_frontend_generator.py` does not run on
python3.6, so it could not be included in the build step for Ubuntu
18.04.

The script `check_datamodels_up_to_date.py` is run as part of `lint.sh`
to ensure that the committed data models are up to date. If the
committed datamodels are out of date, the linting job will fail with the
following type of message:
```
Comparing /host/modules/zivid/settings.py  <-->  /tmp/tmp21nhjawn/settings.py
...
RuntimeError:
Found difference in settings.py.
Please run './continuous-integration/code-generation/docker_generate_datamodels.sh' to regenerate data models.
```


trym-b:
Let's merge https://github.com/zivid/zivid-python/pull/148#pullrequestreview-778220737 first whenever github actions begin operating normally again.

PR #149, opened 2021-10-13
Add Python 3.10 support for Windows

trym-b:


trym-b:
> Test step failed. We may need a newer version of pytest.

Argh, I'll try bumping the dependencies. If it does not work I will close this PR.

PR #156, opened 2022-01-04
Added python 3.10 testing for Linux using the Fedora 35 image

nedrebo:


continuous-integration/linux/platform-dependent/fedora-35/setup.sh
```diff
@@ -0,0 +1,38 @@
...
+dnf --assumeyes install \
+    numactl-libs \
+    redhat-lsb-core \
+    || exit $?
+install_opencl_cpu_runtime || exit $?
+
+function install_www_deb {
+    TMP_DIR=$(mktemp --tmpdir --directory zivid-python-install-www-deb-XXXX) || exit $?
+    pushd $TMP_DIR || exit $?
+    wget -nv "$@" || exit $?
```
trym-b, line 29: Always use the longer options if they exists.

For example
`-r` -> `--requirement`
`-y` -> `--assume-yes`
`-q` -> `--quiet`

trym-b: Same applies for some more instances below.

trym-b (approved):

continuous-integration/linux/platform-dependent/fedora-35/setup.sh
nedrebo: Thx for the review. I agree in principle. However, this script is a verbatim copy of the Fedora 34 script (and probably 33 and 30 also). I do not think it is worth diverting the scripts for this.

PR #160, opened 2022-02-02
Replace `windows-2016` image with `windows-2019`

trym-b:
According to Github `windows-2016` will be deprecated in March 2022. In
order to keep the CI running we should switch to `windows-2019`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,7 +56,7 @@ jobs:
         run: python continuous-integration/notification/notify_teams.py --status ${{ job.status }}
 
   create-windows-binary-distribution:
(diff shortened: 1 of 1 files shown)


eskaur:
Oh this just works now? Nice. Last I tried doing this the build failed :)

eskaur (approved):

trym-b:
> Oh this just works now? Nice. Last I tried doing this the build failed :)

Yeah, perhaps some of the pre-installed dependencies were buggy or something. I tried with `windows-2022` first, but CMake did not find a suitable compiler so I gave up.

RMPR (approved):

Issue #161, opened 2022-02-07
Tests fail on fedora:35 with python3.10


eskaur:
The unit-test "test_capture_from_file" on fedora:35 (python3.10) suddenly started consistently failing. After some investigation we found that it comes from the following error at the time when the file camera is destroyed:
```
pure virtual method called
terminate called without an active exception
Aborted (core dumped)
```

The task is to fix the problem and re-introduce fedora:35/python3.10 into the support matrix.

eskaur:
Archlinux also encountered this issue. Re-enable both fedora35 and archlinux testing when this is fixed.

eskaur:
Solved by https://github.com/zivid/zivid-python/pull/176

PR #162, opened 2022-02-07
Revert "Added python 3.10 testing for Linux using the Fedora 35 image"

eskaur:
The unit-test "test_capture_from_file" on fedora:35 (python3.10)
suddenly started consistently failing. After some investigation it was
found that it comes from the following error at the time when the file
camera is destroyed:

```
pure virtual method called
terminate called without an active exception
Aborted (core dumped)
```

In order to get the CI green again we remove Fedora 35 from the support
matrix for now. The issue https://github.com/zivid/zivid-python/issues/161 was created for fixing this issue and
re-adding it.

This reverts commit https://github.com/zivid/zivid-python/commit/8c703a8e914e564a0204b3d0878e9a3d5e7aea18.


nedrebo:
Can you add a reason to the commit message?

eskaur:
> Can you add a reason to the commit message?

Yeah sorry, should have done that. Done now.

Issue #42, opened 2019-12-06
Enable getting camera intrinsics from zivid-python


SatjaSivcev:
This functionality exists the the C++ SDK. It needs to be enabled for zivid-python as well.

Useful links:

http://www.zivid.com/hubfs/softwarefiles/releases/1.7.0+a115eaa4-4/doc/cpp/classZivid_1_1Camera.html

http://www.zivid.com/hubfs/softwarefiles/releases/1.7.0+a115eaa4-4/doc/cpp/classZivid_1_1CameraIntrinsics.html

ebruun:
is there any update on this?

SatjaSivcev:
Before we implement this feature in zivid-python, you can get camera intrinsic for your particular camera using our C++ or C# SDK since they are constant. You can then save them to file and read them in Python, or just hard code them.

Check out our code samples to get camera intrinsic:
https://github.com/zivid/zivid-cpp-samples/tree/master/source/Camera/InfoUtilOther/GetCameraIntrinsics
https://github.com/zivid/zivid-csharp-samples/tree/master/source/Camera/InfoUtilOther/GetCameraIntrinsics

apockill:
This feature would be a huge nice-to-have when getting started with the zivid. Is there any timeline for this?

SatjaSivcev:
There is no timeline for implementing this feature as of now. Can you share what you want to achieve by using intrinsic? I am asking because there could be an alternative solution, and if there is, I would like to share it.

apockill:
I'm writing my own ROS2 interface for the zivid, and I would like it to work with multiple zivid cameras. I need to create a `CameraInfo` message. Since I don't want to hardcode intrinsics, it would be best if I could request them from the camera.

runenordmo:
Hi, @apockill!

I don't have any further update for this feature than what @SatjaSivcev has mentioned.
But I would like to point to where we construct the `sensor_msgs/CameraInfo` in our zivid-ros wrapper:
https://github.com/zivid/zivid-ros/blob/3be4fbaa0ee7ce1ee796dd29833bd181ad8167fa/zivid_camera/src/zivid_camera.cpp#L649-L695.

Please let us know if you have any more questions.

eskaur:
@ebruun / @apockill : This feature has now been merged to zivid-python. It will be officially release to PyPI in a day or so, but feel free to start using it straight from `master`. :)

ebruun:
@eskaur thanks a lot!

[...]

PR #163, opened 2022-02-07
Add experimental camera intrinsics API

eskaur:
This commit expands the Python wrapper to give access to the
experimental camera intrinsics API. This entails two new functions:
```
from zivid.experimental.calibration import intrinsics
from zivid.experimental.calibration import estimate_intrinsics
```

As part of this the CameraIntrinsics data-model also needs to be
wrapped. Since this data-model (as opposed to the above mentioned
functions) is not technically experimental in the C++ API, it is not
added as experimental here either.

This resolves Issue #42


trym-b (approved):
Looks good to me.

apockill:
Awesome, this works great!

PR #167, opened 2022-02-16
Upgrade to PyBind11 2.9.1

Skylion007:
* This adds supports for mapping std::NestedException to nested equivalent Python Exceptions. I implemented it in PyBind11 and it was mentioned that in the issue that it would be useful for this repo.


nedrebo:
That's awesome, we already noticed that this was implemented. @eskaur is already working on bumping it in https://github.com/zivid/zivid-python/compare/bump-pybind11-version. I think he hit some CI trouble, so I'll leave it to him to sort that out.

But anyway, thx for implementing this in pybind11, this change will improve our error reporting a lot :+1:

PR #169, opened 2022-05-19
Fix ci may2022

eskaur:


trym-b (approved):

apockill:
@eskaur Thank you for being so quick on the ball with these fixes! I was just putting together a reproducible example when I found the problem had already been fixed :grin: :tada:

PR #174, opened 2022-06-07
Update third party licenses

knzivid:
Removes dependency on LGPL bits in eigen. Concatenates all third party licenses.


trym-b (approved):

RMPR (approved):

knzivid:
@eskaur @nedrebo @apartridge Do you have any comments?

eskaur (approved):

nedrebo:
I agree with the LGPL change, nice catch!

But I do not understand all the license files added, or the change seems incomplete.

Maybe I missed it, but it looks like the licenses added here are not embedded in the binary packages exported from the repo, they are just copied into the repo? If that is the case, then I don't see the value added, only the extra maintenance cost to keep the licenses in sync. It is not needed to have licenses in the repos just because a dependency is used, we also need to get them into the artifacts that are distributed.

knzivid:
Adding the licenses to the binary wheels is a nice-to-have. The value added is that one can look up the licenses from README and in my (non-lawyer) opinion, that counts as _attribution_. Before this PR was merged, do you think we attributed these open source dependencies elsewhere?

PR #186, opened 2022-10-07
Expose Zivid::Matrix4x4 to the Python bindings 

Ayxan13:
The Matrix4x4 API is largely similar to that of the native C++ version. But the Matrix4x4 objects are convertible to and from numpy.ndarray:

```
matrix = zivid.Matrix4x4(...)
numpy_array = numpy.array(matrix)
matrix = zivid.Matrix4x4(numpy_array)
```

All our API functions that work with numpy arrays also work with zivid.Matrix4x4. In addition, all numpy functions that expect numpy.ndarray of float32[4, 4] also work with zivid.Matrix4x4.

### DIFF
--- a/modules/zivid/__init__.py
+++ b/modules/zivid/__init__.py
@@ -1,6 +1,5 @@
 """This file imports all non protected classes, modules and packages from the current level."""
 
-
 import zivid._version
 
 __version__ = zivid._version.get_version(__name__)  # pylint: disable=protected-access
@@ -22,3 +21,4 @@ from zivid.settings import Settings
 from zivid.settings_2d import Settings2D
 from zivid.camera_info import CameraInfo
 from zivid.camera_intrinsics import CameraIntrinsics
+from zivid.matrix4x4 import Matrix4x4
--- a/modules/zivid/matrix4x4.py
+++ b/modules/zivid/matrix4x4.py
@@ -1,0 +1,89 @@
+"""Contains Matrix4x4 class."""
+import pathlib
+import _zivid
+
+
+class Matrix4x4(_zivid.Matrix4x4):
+    """Matrix of size 4x4 containing 32-bit floats."""
+
+    def __init__(self, arg=None):
+        """
+        Overloaded constructor.
+
+        Does different kinds of initializations depending on the argument:
+
+        * None or no arguments -> Zero initializes all values.
+        * 1 dimensional List or numpy.ndarray of 16 floats -> Map 1D array of 16 values into this 2D array of 4x4.
+        * 2 dimensional List or numpy.ndarray of 4x4 floats -> Copy 2D array of size 4x4.
+        * str or pathlib.Path -> Load the matrix from a file.
+
+        Args:
+            arg: Any of the above-mentioned.
+        """
+        if arg is None:
+            super().__init__()
+        elif isinstance(arg, pathlib.Path):
+            super().__init__(str(arg))
+        else:
+            super().__init__(arg)
+
+    def inverse(self):
+        """
+        Return the inverse of this matrix.
+
+        An exception is thrown if the matrix is not invertible.
+
+        Returns:
+            A new matrix, holding the inverse
+        """
+        return Matrix4x4(super().inverse())
+
+    def load(self, file_path):
+        """
+        Load the matrix from the given file.
+
+        Args:
+            file_path: path for the file to load the matrix from.
+        """
+        super().load(str(file_path))
+
+    def save(self, file_path):
+        """
+        Save the matrix to the given file.
+
+        Args:
+            file_path: path for the new file to save the matrix in.
+        """
+        super().save(str(file_path))
+
+    def __getitem__(self, indexes):
+        """
+        Access specified element with bounds checking.
+
+        Args:
+            indexes: a tuple of 2 integers as the indexes.
+
+        Returns:
+            The accessed item.
+        """
+        return super()._getitem(indexes)
+
+    def __iter__(self):
+        """
+        Return an iterator to iterate all 16 elements as if this is a 1D array.
+
+        Returns:
+            the iterator.
+        """
+        iterator = super().__iter__()
+        return iterator
+
+    def __setitem__(self, indexes, value):
+        """
+        Set specified element with bounds checking.
+
+        Args:
+            indexes: a tuple of 2 integers as the indexes.
+            value: value to set the specific item to.
+        """
+        super()._setitem(indexes, value)
--- a/samples/sample_calibrate_eye_to_hand.py
+++ b/samples/sample_calibrate_eye_to_hand.py
@@ -1,5 +1,7 @@
 """Hand-eye calibration sample."""
+from pathlib import Path
 import datetime
+import tempfile
 
 import numpy as np
 import zivid
@@ -74,6 +76,14 @@ def _main():
         print("Result:\n{}".format(calibration_result))
     else:
         print("FAILED")
+
+    print("Getting calibration result transformation matrix")
+    transformed_numpy_matrix = calibration_result.transform()
+
+    print("Saving calibration result transformation matrix")
+    with tempfile.TemporaryDirectory() as tempdir:
+        file_path = Path(tempdir) / "hand_eye.yml"
+        zivid.Matrix4x4(transformed_numpy_matrix).save(file_path)
 
 
 if __name__ == "__main__":
--- a/test/calibration/test_hand_eye.py
+++ b/test/calibration/test_hand_eye.py
@@ -11,7 +11,6 @@ def test_handeye_input_init_failure(checkerboard_frames, transform):
 
 
 def test_handeye_input(checkerboard_frames, transform):
-
     import numpy as np
     import zivid.calibration
 
@@ -41,7 +40,6 @@ def test_handeye_input(checkerboard_frames, transform):
 def test_eyetohand_calibration(
     handeye_eth_frames, handeye_eth_poses, handeye_eth_transform
 ):
-
     import numpy as np
     import zivid.calibration
 
@@ -79,3 +77,28 @@ def test_eyetohand_calibration(
         assert residual.translation() >= 0.0
         assert isinstance(residual.rotation(), float)
         assert residual.rotation() >= 0.0
+
+
+def test_eyetohand_calibration_save_load(handeye_eth_frames, handeye_eth_poses):
+    from pathlib import Path
+    import tempfile
+    import zivid
+    import numpy as np
+
+    handeye_output = zivid.calibration.calibrate_eye_to_hand(
+        [
+            zivid.calibration.HandEyeInput(
+                zivid.calibration.Pose(pose_matrix),
+                zivid.calibration.detect_feature_points(frame.point_cloud()),
+            )
+            for frame, pose_matrix in zip(handeye_eth_frames, handeye_eth_poses)
+        ]
+    )
+
+    with tempfile.TemporaryDirectory() as tmpdir, zivid.Application() as _:
+        transform = handeye_output.transform()
+        file_path = Path(tmpdir) / "matrix.yml"
+        zivid.Matrix4x4(transform).save(file_path)
+        np.testing.assert_allclose(
+            zivid.Matrix4x4(transform), zivid.Matrix4x4(file_path), rtol=1e-6
+        )
--- a/test/calibration/test_multi_camera.py
+++ b/test/calibration/test_multi_camera.py
@@ -55,3 +55,25 @@ def test_multicamera_calibration(checkerboard_frames, multicamera_transforms):
             assert residual.translation() == 0.0
         else:
             assert residual.translation() > 0.0
+
+
+def test_multicamera_calibration_save_load(checkerboard_frames):
+    from pathlib import Path
+    import tempfile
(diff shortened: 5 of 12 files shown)


test/test_matrix4x4.py
```diff
@@ -0,0 +1,243 @@
...
+    assert str(matrix) == (
+        "[ [ 0.000000,  1.000000,  2.000000,  3.000000], \n"
+        "  [ 4.000000,  5.000000,  6.000000,  7.000000], \n"
+        "  [ 8.000000,  9.000000,  10.000000,  11.000000], \n"
+        "  [ 12.000000,  13.000000,  14.000000,  15.000000] ]"
+    )
+
+
+if zivid.SDKVersion.major > 2 or (
+    zivid.SDKVersion.major == 2 and zivid.SDKVersion.minor >= 8
```
eskaur, line 204: I don't think this is necessary since this test will not be backported to older SDKs.
Each commit in this repo (when including unit-tests) is compatible with exactly one version of the SDK.

Ayxan13: I did that to pass the CI for now. I am going to remove it (and the one in C++ side) after v2.8 is released. We don't have to merge this PR until then.

eskaur:
Regarding the commit message:

- I think it needs a body describing the API extension in a bit more detail.
- We don't usually reference internal issue tags (ZIVID-XXXX) in this public repo.

test/test_matrix4x4.py
```diff
@@ -0,0 +1,241 @@
...
+
+    assert str(matrix) == (
+        "[ [ 0.000000,  1.000000,  2.000000,  3.000000], \n"
+        "  [ 4.000000,  5.000000,  6.000000,  7.000000], \n"
+        "  [ 8.000000,  9.000000,  10.000000,  11.000000], \n"
+        "  [ 12.000000,  13.000000,  14.000000,  15.000000] ]"
+    )
+
+
+def test_save():
```
eskaur, line 203: Suggestion:
Can you also extend `test_hand_eye.py` and `test_multi_camera.py` to save the results to file using the new Matrix class? Then we have a test that ensures that this stuff is always compatible with the calibration library :)

eskaur: Also, you should demonstrate this new feature in `samples/sample_calibrate_eye_to_hand.py`.

test/test_matrix4x4.py
```diff
@@ -0,0 +1,241 @@
+import os
+import tempfile
+import numpy
+import pytest
+import zivid
+
```
eskaur, line 6: Quite thorough testing. Nice :)

modules/zivid/__init__.py
```diff
@@ -22,3 +21,4 @@
 from zivid.settings_2d import Settings2D
 from zivid.camera_info import CameraInfo
 from zivid.camera_intrinsics import CameraIntrinsics
+from _zivid import Matrix4x4
```
apartridge, line 24: I don't know how the wrapper is designed, but to me it looks strange that only Matrix4x4 has underscore `_zivid` prefix.

test/test_matrix4x4.py
apartridge: Re. this; as I understand it, this Matrix4x4 class can be used as inputs in our various calibration API's (which also should take numpy arrays directly)? If so it would be good to test that also.

test/test_matrix4x4.py
```diff
@@ -0,0 +1,241 @@
...
+        "  [ 8.000000,  9.000000,  10.000000,  11.000000], \n"
+        "  [ 12.000000,  13.000000,  14.000000,  15.000000] ]"
+    )
+
+
+def test_save():
+    with tempfile.TemporaryDirectory() as tmpdir, zivid.Application() as _:
+        file = os.path.join(tmpdir, "matrix_saved.yml")
+        matrix = zivid.Matrix4x4(
+            [[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11], [12, 13, 14, 15]]
```
apartridge, line 207: Could you change some of these values to be floating points? This way we also check that it does handle floating point values.

modules/zivid/__init__.py
apartridge: Ok I see. I think if @eskaur agrees with current approach then it is super. :+1:

test/test_matrix4x4.py
Ayxan13: @apartridge [I've added this line](https://github.com/zivid/zivid-python/blob/c45b27f7ea5fe747ba1b77825199c231495161b6/test/test_matrix4x4.py#L311) which uses the `Pose` constructor with `zivid.Matrix4x4`, which expects a numpy array and the conversion is done implicitly.

samples/sample_calibrate_eye_to_hand.py
```diff
@@ -75,6 +77,14 @@ def _main():
     else:
         print("FAILED")
 
+    print("Transforming calibration result")
+    transformed_numpy_matrix = calibration_result.transform()
+
+    print("Saving transformed calibration result")
+    with tempfile.TemporaryDirectory() as tempdir:
+        file_path = os.path.join(tempdir, "hand_eye.yml")
```
eskaur, line 85: `os.path` is old-school for most things.
The modern way would be `file_path = Path(tempdir) / "hand_eye.yml"`

test/calibration/test_hand_eye.py
```diff
@@ -79,3 +77,28 @@ def test_eyetohand_calibration(
...
+                zivid.calibration.Pose(pose_matrix),
+                zivid.calibration.detect_feature_points(frame.point_cloud()),
+            )
+            for frame, pose_matrix in zip(handeye_eth_frames, handeye_eth_poses)
+        ]
+    )
+
+    with tempfile.TemporaryDirectory() as tmpdir, zivid.Application() as _:
+        transform = handeye_output.transform()
+        file_path = os.path.join(tmpdir, "matrix.yml")
```
eskaur, line 100: Same here. Use pathlib.

test/calibration/test_multi_camera.py
```diff
@@ -55,3 +55,25 @@ def test_multicamera_calibration(checkerboard_frames, multicamera_transforms):
...
+
+    multicamera_output = zivid.calibration.calibrate_multi_camera(
+        [
+            zivid.calibration.detect_feature_points(frame.point_cloud())
+            for frame in checkerboard_frames
+        ]
+    )
+
+    with tempfile.TemporaryDirectory() as tmpdir, zivid.Application() as _:
+        file_path = os.path.join(tmpdir, "matrix.yml")
```
eskaur, line 74: Same here. Use pathlib.

test/test_matrix4x4.py
eskaur: Very nice.

modules/zivid/__init__.py
eskaur: It is fine as long as the customer never has to write **`_zivid`**. That is still the case, right?

eskaur: Actually, @apartridge has a point here. For everything else we have a front-end class with actual Python source-code under `modules/zivid`. This is useful because the source code in that directory also serves as documentation. You class `_zivid.Matrix4x4` does not provide a docstring, for example.

So I think you should make a file `modules/zivid/matrix.py` that implements a class `zivid.Matrix4x4` that is a thin wrapper around `_zivid.Matrix4x4` with docstrings on all methods.

test/test_matrix4x4.py
```diff
@@ -0,0 +1,314 @@
...
+        with open(file, "r", encoding="utf8") as file:
+            assert expected_content in file.read()
+
+
+def test_load():
+    with tempfile.TemporaryDirectory() as tmpdir, zivid.Application() as _:
+        file = os.path.join(tmpdir, "matrix_saved.yml")
+        zivid.Matrix4x4(sample_2d_list()).save(file)
+        matrix = zivid.Matrix4x4()
+        matrix.load(file)
```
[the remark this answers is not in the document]
Ayxan13: It is tested here https://github.com/zivid/zivid-python/blob/b339eb702c741f8ef46d529ace24159087b36362/test/test_matrix4x4.py#L237

modules/zivid/__init__.py
Ayxan13: @eskaur How about I add docstrings in the pybind code as well? It seems like an overkill to create a class just for the docstrings.

eskaur: The philosophy is that the user should not even notice that we are using pybind11. It is an implementation detail.
If they open a sample that uses zivid-python, they should be able to right-click->"Go to source" and end up in readable Python code. If we only expose the pybind11 stuff directly, there is no source code with docstrings to show.

Ayxan13: @eskaur Understandable. I've added a python class to adapt this one, including the docstrings, in `zivid.matrix4x4`.

modules/zivid/matrix4x4.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Args:
+            arg: Any of the above-mentioned.
+        """
+        if arg is None:
+            super().__init__()
+        elif isinstance(arg, pathlib.Path):
+            super().__init__(str(arg))
+        else:
+            super().__init__(arg)
```
eskaur, line 73: I think the constructor should always be the first listed method.

modules/zivid/matrix4x4.py
```diff
@@ -0,0 +1,93 @@
...
+    def save(self, file_path):
+        """
+        Save the matrix to the given file.
+
+        Args:
+            file_path: path for the new file to save the matrix in.
+        """
+        if isinstance(file_path, pathlib.Path):
+            file_path = str(file_path)
+        super().save(file_path)
```
eskaur, line 40: You can just do `super().save(str(file_path))` because `str()` on a string does nothing :)

eskaur: Same in the load() method.

modules/zivid/matrix4x4.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(self, arg=None):
+        """
+        Overloaded constructor.
+
+        Does different kinds of initializations depending on the argument:
+
+        * None or no arguments -> Zero initializes all values.
+        * List[float[16]] -> Map 1D array of 16 values into this 2D array of 4x4.
+        * List[List[float[4]][4]] -> Copy 2D array of size 4x4.
```
eskaur, line 62: This can also be numpy arrays, right?

modules/zivid/matrix4x4.py
Ayxan13: It is possibly something else though. For example, in your sample `matrix.save(123)` would compile because `123` can be convertible to a string. But it should definitely not compile.

modules/zivid/matrix4x4.py
Ayxan13: Yes, definitely. A bit harder to word in here. How about:

```
* ndarray/list of 16 floats -> ...
* ndarray/list of 4x4 float matrix -> ...
```

modules/zivid/__init__.py
eskaur: Nice!

modules/zivid/matrix4x4.py
eskaur: Good point.
But it seems we already are pretty liberal on the string conversions here,
https://github.com/zivid/zivid-python/blob/17ad11301021cb4aac94f09892f86904bd42009d/modules/zivid/application.py#L31

so parhaps do the same here for consistency?

modules/zivid/matrix4x4.py
eskaur: Yeah I like that.

modules/zivid/matrix4x4.py
Ayxan13: Oh, then okay. Consistency is important.

eskaur (approved):
Very nice work!

samples/sample_calibrate_eye_to_hand.py
```diff
@@ -75,6 +77,14 @@ def _main():
     else:
         print("FAILED")
 
+    print("Transforming calibration result")
```
apartridge, line 80: You are not "transforming the calibration result", but instead getting the calibration results transform. So I would update the print to reflect this.

samples/sample_calibrate_eye_to_hand.py
```diff
@@ -75,6 +77,14 @@ def _main():
...
+    transformed_numpy_matrix = calibration_result.transform()
+
+    print("Saving transformed calibration result")
```
apartridge, line 83: Same here, update comment, to fex "Saving calibration result transformation matrix" or similar.

samples/sample_calibrate_eye_to_hand.py
Ayxan13: Updated it to say `"Getting calibration result transformation matrix"`

samples/sample_calibrate_eye_to_hand.py
Ayxan13: Updated it to say `"Saving calibration result transformation matrix"`.

apartridge (approved):

PR #189, opened 2022-10-24
Windows: Ignore TemporaryDirectory errors

knzivid:
It is not a big deal if the cleanup of installers fails, as the job runners are ephemeral. Ignoring it is better than retrying with a backoff and holding a runner for longer than needed.

MISC

---

When necessary, projects like https://github.com/pypa/pip/pull/2397 do implement a retry.


eskaur (dismissed):

knzivid:
Tried backporting just TemporaryDirectory, but it requires more backports. Not a huge fan of backporting more code from python 3.10. Closing the PR for now until I find a better workaround.

PR #191, opened 2022-10-27
Simplify test matrix table

eskaur:
Listing the SDK version in the test matrix is not really useful. Each commit of `zivid-python` supports exactly one version of the SDK, and that version is stated further up in the README anyway. Simplifying this will reduce the manual work of bumping the supported SDK version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,16 +153,16 @@ Please visit [Zivid Knowledge Base][zivid-knowledge-base-url] for general inform
 
 ## Test matrix
 
-| Operating System | Python version            | Zivid SDK version |
-| :--------------- | :------------------------ | :---------------- |
-| Ubuntu 20.04     | 3.8                       | 2.8.0             |
-| Ubuntu 18.04     | 3.6                       | 2.8.0             |
-| Fedora 30        | 3.7                       | 2.8.0             |
-| Fedora 33        | 3.9                       | 2.8.0             |
-| Fedora 34        | 3.9                       | 2.8.0             |
-| Fedora 35        | 3.10                      | 2.8.0             |
-| Windows 10       | 3.7, 3.8, 3.9, 3.10, 3.11 | 2.8.0             |
-| Arch Linux       | latest                    | 2.8.0             |
+| Operating System | Python version            |
+| :--------------- | :------------------------ |
+| Ubuntu 20.04     | 3.8                       |
+| Ubuntu 18.04     | 3.6                       |
+| Fedora 30        | 3.7                       |
+| Fedora 33        | 3.9                       |
+| Fedora 34        | 3.9                       |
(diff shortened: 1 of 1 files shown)


nedrebo (commented):
What you write is not entirely correct.

According to the readme [the wapper does support older SDKs](https://github.com/zivid/zivid-python#install-version-100-of-zivid-python-using-zivid-sdk-version-130).

The only thing potentially stopping you from using the wrapper together with older SDKs is if the wrapper started using API features that are not in the older SDK.

This table is, as I understand it, not about what is possible, but about what we actually test.

----

Related, looking at this, I also think the `Arch Linux` line is wrong. That test is against whatever is released, not against any pinned version since it uses AUR, so it should really be `latest`.

The same may be true for the other entries, depending on if they use a hard-coded version or if they dynamically fetch the latest version from zivid.com.

nedrebo:
> However, If I merely check out `master` branch and do `pip install .`,

Installing from repo is not recommend in the readme, it recommends `pip install zivid`

> To support other SDKs you have to fiddle with setup.py.

The readme say that `pip install zivid=1.2.3.4.5.6` works to select SDK=4.5.6, is that not true?

> What if we between the Test matrix and the table put This version of zivid-python is tested against Zivid SDK 2.8.0?

I'm fine with that. Given, that is is true. You did not answer if this is dynamic or hard-coded. Do you know?

For Arch Linux at least, we do test against the latest released SDK. If we, for some reason, do not update the wrapper immediately after a release, the readme will be (and is in the current version) wrong.

nedrebo:
> If I merely check out `master` branch and do `pip install .`,

I don't think this is "merely". I think most users get this from PyPi and never visit GitHub.

PyPi also displays the Readme, so it must also (and even more importantly) make sense there.

eskaur:
I guess it depends on this question: Does the Readme.md on a commit represent just that commit, or all previous releases?

> Installing from repo is not recommend in the readme, it recommends pip install zivid

The readme actually mentions both ways. There's one header for each.

> The readme say that pip install zivid=1.2.3.4.5.6 works to select SDK=4.5.6, is that not true?

That is true.

> I'm fine with that. Given, that is is true. You did not answer if this is dynamic or hard-coded. Do you know?

The SDK version tested in the CI is hard-coded in the repo with explicit links to installers.

> For Arch Linux at least, we do test against the latest released SDK.

This is not actually true. There is a hardcoded reference to a specific version.

nedrebo (approved):
Thx for clearing up all the points.

The 1:1 mapping between the wrapper and the SDK is apparently harder than I remembered. And, it is probably also the correct tradeoff to keep it this interlocked when considering flexibility vs. maintenance work.

I support this simplification :)

PR #192, opened 2022-10-27
Specify which Python versions are supported

torbsorb:
Clarify exactly which Python versions are supported, and add the classifiers that makes it clear in the sidebar on pypi.org.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Zivid Python is the official Python package for Zivid 3D cameras. Read more abou
 
 ### Dependencies
 
-* [Python](https://www.python.org/) version 3.6 or higher
+* [Python](https://www.python.org/) version 3.6, 3.7, 3.8, 3.9 and 3.10
 * [Zivid SDK][zivid-download-software-url] version 2.8.0 (see [here][zivid-software-installation-url] for help)
 * [Compiler](doc/CompilerInstallation.md) with C++17 support
 
--- a/setup.py
+++ b/setup.py
@@ -139,7 +139,11 @@ def _main():
         ],
         classifiers=[
             "License :: OSI Approved :: BSD License",
-            "Programming Language :: Python :: 3",
+            "Programming Language :: Python :: 3.6",
+            "Programming Language :: Python :: 3.7",
+            "Programming Language :: Python :: 3.8",
+            "Programming Language :: Python :: 3.9",
+            "Programming Language :: Python :: 3.10",
             "Programming Language :: Python :: Implementation :: CPython",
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -15,7 +15,7 @@ Zivid Python is the official Python package for Zivid 3D cameras. Read more abou
 
 ### Dependencies
 
-* [Python](https://www.python.org/) version 3.6 or higher
+* [Python](https://www.python.org/) version 3.6, 3.7, 3.8, 3.9 and 3.10
```
eskaur, line 18: Slight complication to consider:
- We don't support 3.6 on Windows anymore (in the sense that we don't build, test & upload a .whl file for that).  This is because 3.6 has been end-of-life for a while now.
- However, we do test Python 3.6 on Linux, because that's the default Python on Ubuntu 18.04 and we still support that.

eskaur: Also we support `3.11` on `master` now. Just hasn't been released to PyPI yet. So 3.11 should be listed here.

PyPI displays a snapshot of the Readme on the commit that is released.

eskaur: Opinions, @nedrebo ?
Should this list show the versions that we support somewhere, or just the versions that we support everywhere?

eskaur: Let's just do the union. Should now be "3.6, 3.7, 3.8, 3.9, 3.10 and 3.11"

trym-b: I think we should just drop 3.6 as it is already been EOL for over a year. Even 3.7 is almost EOL too

eskaur: It's the default Python on Ubuntu 18.04, so I suppose we should drop it when U18 is dropped.

nedrebo: Let's do the ones that are fully supported. I.e that `pip install` will work for on the listed versions.

Drop 3.6 here, but keep testing on 3.6 to keep undocumented support best effort until we hit a real blocker for that support or until Zivid SDK drops support for 18.04.

setup.py
```diff
@@ -139,7 +139,11 @@ def _main():
         ],
         classifiers=[
             "License :: OSI Approved :: BSD License",
-            "Programming Language :: Python :: 3",
+            "Programming Language :: Python :: 3.6",
```
nedrebo, line 142: Drop 3.6 and add 3.11 here also.

SatjaSivcev:
Not sure what I want with this, but have in mind that we also state which python we support here https://support.zivid.com/en/latest/api-reference/sdk-lifecycle.html#python depending on the OS. Take it into consideration when deciding what is the best for this PR, to avoid any confusion.

PR #195, opened 2022-11-30
Allow construction of Matrix4x4 directly from Pose

eskaur:
Functions such as `DetectionResult.pose()` return a `Pose`, not a numpy array. The user may want to save this pose to file, which could only be done by:
```
zivid.Matrix4x4(pose.to_matrix()).save("mypose.yml")
```

This commit makes the process easier by enabling:
```
zivid.Matrix4x4(pose).save("mypose.yml")
```

### DIFF
--- a/modules/zivid/matrix4x4.py
+++ b/modules/zivid/matrix4x4.py
@@ -1,6 +1,7 @@
 """Contains Matrix4x4 class."""
 import pathlib
 import _zivid
+from zivid.calibration import Pose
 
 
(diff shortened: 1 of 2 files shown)


apartridge:
If we change this then we should also change the C++ API to make a Matrix4x4 ctor that takes a pose, for consistency? (Also .NET possibly)? And I am not sure if that is correct thing to do since Matrix4x4 is a more basic primitive than a Pose, so Matrix should not know too much about Pose's?

The first code line (existing code) did not feel like too much coding?

PR #204, opened 2023-02-22
Pin conan version

dragly:
Conan 2.0 was released today and breaks zivid-python with the following error message due to a change in Conan's API:

```
conan install: error: unrecognized arguments: --install-folder
```

This change pins Conan to 1.x and fixes the issue.


eskaur:
@dragly : I think you need to make a branch in this repo (not from a fork) in order to trigger the CI.

PR #207, opened 2023-03-06
De-duplicate exact SDK version used in CI

eskaur:
Version strings like `2.8.1+dd4dffea-1` are used to choose which SDK installers are downloaded in the CI worflows. These were unnecessarily duplicated across many files. This commit defines them in a single place, so that future commits that bumps SDK version can be simpler.


nedrebo (approved):
Awesome :+1:

PR #209, opened 2023-03-23
Bump to sdk 2.9.0

eskaur:
Draft to get early feedback.
One of the commits is temporary hack to make it build against the preview installers.


test/test_region_of_interest.py
```diff
@@ -0,0 +1,56 @@
...
+    settings.region_of_interest.box.extents = [-500.0, 500.0]
+    settings.region_of_interest.box.point_o = [x_min, y_max, 500.0]
+    settings.region_of_interest.box.point_a = [x_max, y_max, 500.0]
+    settings.region_of_interest.box.point_b = [x_min, y_min, 500.0]
+    with file_camera.capture(settings) as frame_roi:
+        xyz_roi = frame_roi.point_cloud().copy_data("xyz")
+    assert np.nanmin(xyz_roi[:, :, 0]) >= x_min
+    assert np.nanmax(xyz_roi[:, :, 0]) <= x_max
+    assert np.nanmin(xyz_roi[:, :, 1]) >= y_min
+    assert np.nanmax(xyz_roi[:, :, 1]) <= y_max
```
eskaur, line 56: If you just wonder what the front-end ROI API looks like, this is the place to look.

src/DataModel.cpp
```diff
@@ -1,16 +1,83 @@
...
+                 [](const R &self) {
+                     return std::array<double, 2>{ self.min(), self.max() };
+                 })
+            .def("__repr__", &R::toString)
+            .def("to_string", &R::toString)
+            .def(pybind11::self == pybind11::self) // NOLINT
+            .def(pybind11::self != pybind11::self) // NOLINT
+            .def("is_in_range", &R::isInRange);
+
+        pybind11::implicitly_convertible<pybind11::tuple, R>();
```
knzivid, line 38: I tried and failed getting `pybind11::implicitly_convertible<pybind11::iterable, R>();` to work. If this works, it would cover all three and more (like numpy arrays). Any idea why?

src/include/ZividPython/DataModelWrapper.h
```diff
@@ -230,6 +254,9 @@ namespace ZividPython
                     pyClass.def(py::self < py::self); // NOLINT
                 }
 
+                pyClass.attr("is_array_like") = TypeIsZividRange<typename Target::ValueType>::value
```
knzivid, line 257: I rabbit-holed myself into attributes vs properties in python because I don't use attrs often. TIL while an attribute can be set by the user (`obj.is_array_like = False`), a read-only property can not be. Not asking to change it here, of course.

src/DataModel.cpp
```diff
@@ -1,16 +1,83 @@
...
 namespace ZividPython::DataModel
 {
     void wrapAsSubmodule(pybind11::module &dest)
     {
         using namespace Zivid::DataModel;
         ZIVID_PYTHON_WRAP_ENUM_CLASS(dest, NodeType);
+
+        using Range = Zivid::Range<double>;
+        using PointXYZ = Zivid::PointXYZ;
+        ZIVID_PYTHON_WRAP_CLASS(dest, Range);
```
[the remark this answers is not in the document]
eskaur: We are free to change this in the future since this pybind11 generated class is not part of the public zivid-python API at all.
But I could rename it to `RangeDouble` or something if you think that reduces confusion.

test/test_region_of_interest.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_region_of_interest_depth(file_camera):
+    import zivid
+
+    z_min = 600.0
+    z_max = 625.0
+
+    # Make sure default capture has points outside the range we want
+    settings = zivid.Settings(acquisitions=[zivid.Settings.Acquisition()])
```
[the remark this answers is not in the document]
eskaur: Good question. It works because of the clever fixture system we have here. This test depends on the `"file_camera"` fixture, which depends on the `"application"` fixture. The fixtures keep these things alive with a yield statement:

```
@pytest.fixture(name="application", scope="module")
def application_fixture():
    with zivid.Application() as app:
        yield app
```

src/DataModel.cpp
eskaur: I tried that here and it seems to work (it builds and all of the tests pass). Pushed fixup now.

Regarding why you failed to do something similar elsewhere: Maybe you were missing some includes? `#include <pybind11/pybind11.h>` is not always sufficient.

knzivid: Nice! Not sure what I messed up, probably includes like you said. I don't think I included anything extra.

src/DataModel.cpp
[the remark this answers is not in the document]
knzivid: I don't have a strong preference. As long as we can break the internal names, I am happy with just Range :)

continuous-integration/code-generation/datamodel_frontend_generator.py
```diff
@@ -518,6 +532,7 @@ def _parse_internal_datamodel(current_class: Any) -> NodeData:
         snake_case=_to_snake_case(current_class.name),
         is_leaf=is_leaf,
         is_enum=is_enum_class,
+        is_array_like=is_leaf and current_class.is_array_like,
```
knzivid, line 535: is there a situation where `current_class.is_array_like` is true and `is_leaf` false?

apartridge: @knzivid Eskil is away for 3 weeks, do you think this is a blocker for merging this?

knzivid: No, not if is_array_like in C++ is set to false for non leaf nodes? I am not sure what _array_ here implies, so this suggestion could be wrong.

knzivid: Figured why this exists - this is equivalent to `hasattr(current_class, "is_array_like") and current_class.is_array_like`. The `is_leaf` check assumes the C++ implementation detail that `is_array_like` is only set for leaf nodes. I propose using hasattr or adding a comment about this implementation detail.

knzivid: added comment

knzivid (dismissed):
approving the first two commits. waiting for approval from @apartridge for the third one

apartridge (approved):
cf4837a8077566f9890f897fa09243dbc4e71c4b looks good.

apartridge (approved):

PR #211, opened 2023-04-21
Add `Frame2D.camera_info` property

eskaur:
In SDK 2.9 `Frame2D` was extended with a `CameraInfo` getter, similar to what `Frame` already had. This commit exposes this to zivid-python.

### DIFF
--- a/test/test_frame_2d.py
+++ b/test/test_frame_2d.py
@@ -38,6 +38,15 @@ def test_info(physical_camera_frame_2d):
 
 
 @pytest.mark.physical_camera
+def test_camera_info(physical_camera_frame_2d):
+    from zivid.camera_info import CameraInfo
+
+    camera_info = physical_camera_frame_2d.camera_info
+    assert camera_info
+    assert isinstance(camera_info, CameraInfo)
+
+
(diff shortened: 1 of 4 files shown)


test/test_frame_2d.py
```diff
@@ -37,6 +37,15 @@ def test_info(physical_camera_frame_2d):
     assert isinstance(info, zivid.FrameInfo)
 
 
+@pytest.mark.physical_camera
+def test_camera_info(physical_camera_frame_2d):
+    from zivid.camera_info import CameraInfo
+
+    camera_info = physical_camera_frame_2d.camera_info
+    assert camera_info
+    assert isinstance(camera_info, CameraInfo)
```
apartridge, line 46: Looks OK, I would have liked some more testing of the properties of the CameraInfo, such as the model and serialnumber. I see also test_frame.py has the same test though, so then both could be improved a bit.

apartridge (approved):

test/test_frame_2d.py
eskaur: The actual properties are tested in `test/test_camera_info.py`.

PR #214, opened 2023-05-04
Fix broken CI

knzivid:
Archlinux moving to gcc 13 triggered a series of cascading issues


nedrebo (commented):
I am not sure this will make `pip install zivid` work on Arch.

I don't know much about Conan or how the toolchain is set up in various scenarios. However, quickly looking at the patch, it seems we now introduce some special CI setup, the `settins.yml` file, but I cannot see that it is embedded in the source distribution. My guess is that source distribution will not work on Arch, even after this change. Any thoughts on that?

@eskaur, do we test the source distributions on vanilla OS'es?

Even though we do not have coverage of this, it is an aim for this project to work on any "normal" distro. We know people use this on many OSes and more and more people will have GCC 13 soon.

The solution we select here should be one that will work for any OS with GCC 13.

knzivid:
You are right, the source distribution is broken on arch. This PR makes the CI green and does not fix the customer situation. I have a follow-up in mind to precisely control the conan user home and override it from setup.py but my first goal is to make zivid-python green. If possible, I would remove conan completely from zivid-python because it is a moving target that would break again. 

When a customer installs zivid-python on a new enough distro that uses gcc 13 or if clang releases a new version that's not tracked in conan's settings.yml, zivid-python will break. This is a pitfall on all platforms.

As the CI triage person for this sprint, I can only take detours into fixing red pipelines if they are simple enough. If you don't accept this PR, I will have to close it and I can't guarantee I will get back to the perfect solution soon.

nedrebo:
I IRL'ed with @eskaur.

The problem here, as far as I understand it, is that Arch moved more quicker than Conan. And this is probably a transient error. However, one that appears now and then in various forms. It can also be arch is to new for some pip packages.

We agreed that Arch testing is not the important thing here, and it is not worth the random breakages.

What is important is that we test on a variety of OSes, and not all Ubuntus to make sure we do not have Ubuntu-specific code.

Instead of this PR, @eskaur will do the following within a day or two. Before that it is ok that SDK team ignores CI errors from this repository.

* Remove all Arch testing and code
* Port linting code to the latest released Ubuntu (now 23.04). Move this forward every Ubuntu release.
* Extend test matrices to include all Ubuntu LTS versions supported by Zivid, union 4 latest Ubuntu versions. Move this forward every Ubuntu release.
* Update the test matrix to include the 3 latest Fedora releases. Move this forward every Fedora release.
* Put a doc string over the test matrix to explain which version we should test on.

knzivid:
I agree with the other points but not with the claim that this is transient. This is systematic and will break if a customer uses a different compiler as the system default irrespective of the distro.

Happy to discuss it irl

PR #215, opened 2023-05-05
Stop running Arch in CI, and update test matrix

eskaur:


trym-b (dismissed):
Added one comment; otherwise looks good to me.

.github/workflows/main.yml
```diff
@@ -92,17 +93,20 @@ jobs:
   test-linux-source-distribution:
     needs: create-source-distribution
     runs-on: ubuntu-latest
+    # Build & test job should be run on:
+    # - All Ubuntu LTS supported by Zivid
+    # - All Ubuntu versions newer than the latest LTS
```
nedrebo, line 98: Is it hard to change this to the following?

```suggestion
    # - 4 latest Ubuntu versions
```

The reason I ask is that it is highly likely that we got users on any recent Ubuntu, we do not want to leave them behind immediately when a new LTS is out.

.github/workflows/main.yml
```diff
@@ -92,17 +93,20 @@ jobs:
   test-linux-source-distribution:
     needs: create-source-distribution
     runs-on: ubuntu-latest
+    # Build & test job should be run on:
+    # - All Ubuntu LTS supported by Zivid
```
nedrebo, line 97:
```suggestion
    # - All Ubuntu versions supported by Zivid SDK
```

Technically the same, but more robust formulation without the LTS.

.github/workflows/main.yml
eskaur: I'll try. Could be easy.

.github/workflows/main.yml
```diff
@@ -93,16 +93,20 @@ jobs:
...
+    # - All Ubuntu versions newer than the latest LTS
+    # - The three latest Fedora releases that we can support
     strategy:
       matrix:
         os:
           - ubuntu:18.04
           - ubuntu:20.04
           - ubuntu:22.04
-          - fedora:30
```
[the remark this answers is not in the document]
eskaur: Isn't it still true that the installer still only officially supports 16.04? I haven't changed out the installer.

.github/workflows/main.yml
eskaur: Wait Ubuntu 21.04 and 21.10 are already past end-of-life. The apt repos aren't even up anymore.

nedrebo: Ok, resolving this thread then.

.github/workflows/main.yml
nedrebo: This is a minor comment. Approving now, fix before merging is you agree.

nedrebo (approved):

.github/workflows/main.yml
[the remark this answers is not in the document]
trym-b: Ok, then let's consider it as a follow-up. Might be we do not want to touch this at all

.github/workflows/main.yml
eskaur: Done

trym-b (approved):

PR #217, opened 2023-05-11
MISC: Add support for frame2d.image_rgba()

torbsorb:
We have samples which utilize the new feature where you can directly copy in BGR format. To keep samples aligned we add the missing feature here.

`test_release.py` does not require explicit testing with image_bgra().

pytest has been run with and without the marker `physical_camera`. The test was run on Windows, and the samples tests failed because of something unrelated (not root caused, but points towards `pytest.helpers.run_sample`). All samples were tested directly without pytest. They all pass.

Note! We should also consider adding BGR for pointCloud.

MISC


README.md
```diff
@@ -87,6 +87,7 @@ Data can easily be accessed in the form of Numpy arrays:
     frame = camera.capture(settings)
     xyz = frame.point_cloud().copy_data("xyz") # Get point coordinates as [Height,Width,3] float array
     rgba = frame.point_cloud().copy_data("rgba") # Get point colors as [Height,Width,4] uint8 array
+    bgra = frame.point_cloud().copy_data("bgra") # Get point colors as [Height,Width,4] uint8 array
```
eskaur, line 90: This is misleading because you have not actually added `frame.point_cloud().copy_data("bgra")` in this PR.

I think the right move is to keep this line, and then actually add `frame.point_cloud().copy_data("bgra")` to this PR.

eskaur: Generally I think it would be strange to only add BGRA to 2D capture but not to 3D captures.

Let me know when that is added too and I'll review the rest.

test/test_frame_2d.py
```diff
@@ -5,18 +5,26 @@
...
-    image = physical_camera_frame_2d.image_rgba()
-    assert image is not None
-    assert isinstance(image, zivid.Image)
+    image_rgba = physical_camera_frame_2d.image_rgba()
+    assert image_rgba is not None
+    assert isinstance(image_rgba, zivid.Image)
+
+    image_bgra = physical_camera_frame_2d.image_bgra()
+    assert image_bgra is not None
+    assert isinstance(image_bgra, zivid.Image)
```
eskaur, line 27: Can you also add some sort of test that checks that the RGBA and BGRA from the same Frame2D contain the same image?

Something like (pseudocode):
```
rgba.r == bgra.r
rgba.g == bgra.g
rgba.b == bgra.b
rgba.a == bgra.a
```

eskaur: Same for the frame.point_cloud data when you add that.

src/ReleasableArray2D.cpp
```diff
@@ -71,6 +71,19 @@ namespace ZividPython
             .def_buffer(pointCloudDataBuffer<NativeType, WrapperType, 4>);
     }
 
+    void wrapClass(pybind11::class_<ReleasableArray2D<Zivid::ColorBGRA>> pyClass)
+    {
+        using WrapperType = uint8_t;
+        using NativeType = Zivid::ColorBGRA;
+        static_assert(std::is_same_v<WrapperType, decltype(NativeType::g)>);
+        static_assert(std::is_same_v<WrapperType, decltype(NativeType::g)>);
```
eskaur, line 79: Duplicate line. First one should probably be "b"?

torbsorb: good catch

test/test_frame_2d.py
```diff
@@ -1,22 +1,44 @@
...
+
+
+@pytest.mark.physical_camera
+def test_image_rgba_bgra_correspondence(physical_camera_frame_2d):
+    import zivid
```
eskaur, line 33: Unused import

test/test_frame_2d.py
```diff
@@ -1,22 +1,44 @@
...
+        assert image_rgba is not None
+        assert isinstance(image_rgba, zivid.Image)
+
+    with physical_camera_frame_2d.image_bgra() as image_bgra:
+        assert image_bgra is not None
+        assert isinstance(image_bgra, zivid.Image)
 
 
 @pytest.mark.physical_camera
 def test_image(physical_camera_frame_2d):
```
eskaur, line 19: You ran through all these tests with a physical camera right? 
(important since they are not run in the CI)

test/test_frame_2d.py
```diff
@@ -1,22 +1,44 @@
...
+def test_image_rgba_bgra_correspondence(physical_camera_frame_2d):
+    import zivid
+
+    rgba = physical_camera_frame_2d.image_rgba().copy_data()
+    bgra = physical_camera_frame_2d.image_bgra().copy_data()
+
+    np.testing.assert_array_equal(bgra[:, :, 0], rgba[:, :, 2])
+    np.testing.assert_array_equal(bgra[:, :, 1], rgba[:, :, 1])
+    np.testing.assert_array_equal(bgra[:, :, 2], rgba[:, :, 0])
+    np.testing.assert_array_equal(bgra[:, :, 3], rgba[:, :, 3])
```
eskaur, line 41: Nice, thanks

modules/zivid/image.py
```diff
@@ -18,10 +18,10 @@ def __init__(self, impl):
         Raises:
             TypeError: If argument does not match the expected internal class.
         """
-        if not isinstance(impl, _zivid.ImageRGBA):
+        if not isinstance(impl, _zivid.ImageRGBA | _zivid.ImageBGRA):
```
eskaur, line 21: This syntax was introduced in Python 3.10. Unfortunately we still have to support older Python versions still, where this gives:
```
TypeError: unsupported operand type(s) for |: 'pybind11_type' and 'pybind11_type'
```

While we do test all supported Python versions in the CI, there is a hole in our test coverage: We do not test 2D capture stuff, because that does not work properly with file cameras. I suppose we can do some smoke-tests anyway, but I can deal with that.

For now you have to use the legacy syntax:
```
isinstance(impl, (_zivid.ImageRGBA, _zivid.ImageBGRA))
```

test/test_frame_2d.py
eskaur: If you also remove the empty line below everything should pass CI :)

eskaur:
Looks good. Feel free to rebase on master whenever.

test/test_frame_2d.py
torbsorb: I did not run all again because I didn't have access to physical camera. I'll ask someone for assistance.

eskaur (approved):
Very nice :neckbeard: :+1:

eskaur (approved):

PR #222, opened 2023-05-19
Capture: Unlock GIL during capture to allow async

Ayxan13:
This unlocks GIL during captures so they can be executed in parallel. This also adds a sample to show how to execute captures in parallel.

Implements ZIVID-6981

### DIFF
--- a/samples/sample_capture_async.py
+++ b/samples/sample_capture_async.py
@@ -1,0 +1,65 @@
+"""Sample demonstrating how to capture with multiple cameras at the same time."""
+import concurrent.futures
+import datetime
+import time
+
+import zivid
+
+
+def _capture_sync(cameras: list[zivid.Camera]) -> list[zivid.Frame]:
+    return [
+        camera.capture(
+            zivid.Settings(
+                acquisitions=[
+                    zivid.Settings.Acquisition(
+                        exposure_time=datetime.timedelta(microseconds=100000)
+                    )
+                ]
+            )
+        )
+        for camera in cameras
+    ]
+
+
+def _capture_async(cameras: list[zivid.Camera]) -> list[zivid.Frame]:
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        futures = [
+            executor.submit(
+                camera.capture,
+                zivid.Settings(
+                    acquisitions=[
+                        zivid.Settings.Acquisition(
+                            exposure_time=datetime.timedelta(microseconds=100000)
+                        )
+                    ]
+                ),
+            )
+            for camera in cameras
+        ]
+
+        return [future.result() for future in futures]
+
+
+def _main():
+    app = zivid.Application()
+    cameras = app.cameras()
+    for camera in cameras:
+        camera.connect()
+
+    start = time.monotonic()
+    _capture_async(cameras)
+    end = time.monotonic()
+    print(
+        f"Time taken to capture asynchronously from {len(cameras)} camera(s): {end-start} seconds"
+    )
+
+    start = time.monotonic()
+    _capture_sync(cameras)
+    end = time.monotonic()
+    print(
+        f"Time taken to capture synchronously from {len(cameras)} camera(s): {end-start} seconds"
+    )
+
+
+if __name__ == "__main__":
+    _main()
--- a/src/include/ZividPython/Releasable.h
+++ b/src/include/ZividPython/Releasable.h
@@ -2,12 +2,19 @@
 
 #include <memory>
 #include <optional>
+#include <pybind11/pybind11.h>
 #include <stdexcept>
+
+#define WITH_GIL_UNLOCKED(...)                                                                                         \
+    [&, this] {                                                                                                        \
+        pybind11::gil_scoped_release gilLock;                                                                          \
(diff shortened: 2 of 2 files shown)


samples/sample_capture_async.py
```diff
@@ -0,0 +1,61 @@
+"""Sample demonstrating how to capture with multiple cameras at the same time."""
+import concurrent.futures
+import datetime
+import time
+
+import zivid
+
+
+def _sync_version(cameras: list[zivid.Camera]) -> list[zivid.Frame]:
```
apartridge, line 9: Maybe nicer to name these `_capture_async` and `_capture_sync` or maybe `_(a)sync_capture`

samples/sample_capture_async.py
```diff
@@ -0,0 +1,61 @@
...
+def _main():
+    app = zivid.Application()
+    cameras = app.cameras()
+    for camera in cameras:
+        camera.connect()
+
+    start = time.monotonic()
+    _async_version(cameras)
+    end = time.monotonic()
+    print("Async time:", end - start, "seconds")
```
apartridge, line 52: This text is a bit lacking of context, maybe it can say "Time taken to capture asynchronously from X cameras: Y seconds".

samples/sample_capture_async.py
Ayxan13: Amended

samples/sample_capture_async.py
Ayxan13: Amended

src/include/ZividPython/Releasable.h
```diff
@@ -55,6 +56,13 @@
         return returnType{ impl().functionName(arg1Name) };                                                            \
     }
 
+#define ZIVID_PYTHON_FORWARD_1_ARGS_WRAP_RETURN_GIL_UNLOCKED(returnType, functionName, arg1Type, arg1Name)             \
+    auto functionName(arg1Type arg1Name)                                                                               \
+    {                                                                                                                  \
+        pybind11::gil_scoped_release gilLock;                                                                          \
```
nedrebo, line 62: Idk much about dirty businesses with the GIL. However, since it is opt-in to disable it there must be some downsides. If not we could just add this line to all our macros instead of adding these special macros.

Can you extend the commit message to explain:

* What happens when we disable the GIL temporarily
* What the downsides are
* Why the downsides don't apply in this case
  * Or why the downsides are acceptable in this case

If we need to use these new macros with care also add a doc string per macro.

nedrebo (commented):
The code looks good, but requested a bit more documentation.

nedrebo (commented):
The code looks good, but requested a bit more documentation.

src/include/ZividPython/Releasable.h
Ayxan13: @nedrebo Relevant excerpts from the [pybind11 website](https://pybind11.readthedocs.io/en/stable/advanced/misc.html):

> The Python C API dictates that the Global Interpreter Lock (GIL) must always be held by the current thread to safely access Python objects. As a result, when Python calls into C++ via pybind11 the GIL must be held, and pybind11 will never implicitly release the GIL.

This means pybind11 is being cautious to leave the GIL locked just in case we make any calls to any python functions/objects in our C++ code. But our C++ SDK does not access any Python objects, so it is perfectly safe to unlock the GIL for the duration of the function calls *to our SDK functions*. I don't see any down-sides to this at all.

> pybind11 will ensure that the GIL is held when it knows that it is calling Python code. For example, if a Python callback is passed to C++ code via std::function, when C++ code calls the function the built-in wrapper will acquire the GIL before calling the Python callback. Similarly, the PYBIND11_OVERRIDE family of macros will acquire the GIL before calling back into Python.
>
> When writing C++ code that is called from other C++ code, if that code accesses Python state, it must explicitly acquire and release the GIL.
>
> The classes gil_scoped_release and gil_scoped_acquire can be used to acquire and release the global interpreter lock in the body of a C++ function call. In this way, long-running C++ code can be parallelized using multiple Python threads, but great care must be taken when any gil_scoped_release appear: if there is any way that the C++ code can access Python objects, gil_scoped_acquire should be used to reacquire the GIL.

To make sure that if there is ever a python object in our *wrapper*, we make sure to unlock the GIL only for the duration of the function calls that go to the SDK. There reason it's done for this function, not elsewhere, is that it was needed here specifically. But in there is no reason to not do it for other blocking functions like `connect` and `disconnect`.

nedrebo: Then I suggest KISS, just unlocking GIL in all the existing macros and postponing introducing a more complicated system until the day Zivid SDK actually calls into Python. That is unlikely to happen any time soon.

Thoughts @apartridge?

apartridge: Sounds good to me, I think we just have to be a bit dilligent with testing. We have quite good unit test coverage of the python wrapper, maybe also run the samples and see that it works fine?

Ayxan13: @nedrebo @apartridge I've extended it to all wrapper macros.

nedrebo (approved):

apartridge (approved):

PR #227, opened 2023-06-29
Use temporary out-of-source build directory to enable repeated builds

eskaur:
Currently, running `pip install .` in the repo root cannot be done multiple times. When running the second time, the developer is faced with a large error message. The solution is to delete the `_skbuild` directory that has popped up in the repo root, but the error message gives no hint that this may be the solution. This problem has confused both users and internal developers countless times.

This commit enabled easy repeated builds by overriding the build directory from the repo root to a random temporary directory. The way this is done is arguably a bit monkey-patchy, but to my knowledge there is no other way. This may be better in the modern in-progress re-implementation of `scikit-build` called `scikit-build-core`, which we could migrate to in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,9 +2,10 @@ import tempfile
 import platform
 import subprocess
 from sys import version_info
+from tempfile import TemporaryDirectory
 from pathlib import Path
 from pkgutil import iter_modules
-from skbuild import setup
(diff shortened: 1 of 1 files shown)


trym-b (approved):
Looks good to me :)

apartridge (approved):

PR #229, opened 2023-07-17
Add support for overloads of calibration.intrinsics

torbsorb:
SDK 2.10.0 introduces two overloads of the calibration.intrinsics API. One with settings and one with settings_2d. This commit adds support in the Python wrapper.


Ayxan13 (approved):
Looks good to me!

eskaur (approved):

PR #234, opened 2023-08-15
Bump to sdk 2.10.0

eskaur:


src/Projection.cpp
```diff
@@ -0,0 +1,73 @@
...
+                const auto channels = info.shape[2];
+
+                if(channels != 4)
+                {
+                    throw std::runtime_error("Input image array must have four color channels (BGRA).");
+                }
+
+                const auto resolution = Zivid::Resolution{ static_cast<size_t>(width), static_cast<size_t>(height) };
+                const auto *dataPtr = reinterpret_cast<const Zivid::ColorBGRA *>(imageBGRA.data());
+                const Zivid::Image<Zivid::ColorBGRA> zividImage{ resolution, dataPtr, dataPtr + resolution.size() };
```
apartridge, line 43: It would be better to use the Image constructor that takes unsigned char begin/end ptrs, to avoid strict aliasing rule problems (this constructor was added quite late in 2.10 development so it likely didnt exist when you wrote this).

eskaur: Ah yes, I'll give it a shot.

test/test_projection.py
```diff
@@ -0,0 +1,82 @@
...
+    points = [[0.0, 0.0, 1000.0], [10.0, -10.0, 1200.0]]
+    projector_coords = pixels_from_3d_points(camera=physical_camera, points=points)
+    assert isinstance(projector_coords, list)
+    assert len(projector_coords) == len(points)
+    for coord in projector_coords:
+        assert isinstance(coord, list)
+        assert len(coord) == 2
+        assert isinstance(coord[0], float)
+        assert isinstance(coord[1], float)
+
```
apartridge, line 80: I think it here would be good also to test the actual returned points. I hope they would be consistent, since I assume this is a file camera? Or is this a manual test run on actual camera? By testing the output we know that we have not mixed up the coordinate ordes or similar in the conversion between python and C++.

modules/zivid/experimental/projection.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Returns:
+        The resolution as a tuple (height,width).
+    """
+    return _zivid.projection.projector_resolution(
+        camera._Camera__impl  # pylint: disable=protected-access
+    )
+
+
+def show_image(camera, image_bgra):
```
apartridge, line 104: In the future in the C++ API we may add overloads for rgba, rgb etc. Then we will get a bit hurt here, since there is no type information passed along with image_bgra. Anyway that is a problem for later.

modules/zivid/experimental/projection.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+    )
+
+
+def pixels_from_3d_points(camera, points):
+    """Get 2D projector pixel coordinates corresponding to 3D points relative to the camera.
+
+    This function takes 3D points in the camera's reference frame and converts them to the
+    projector's 2D (XY) frame using the internal calibration of a Zivid camera. The returned
+    points maps to projector pixels as round(X)->col and round(Y)->row.
```
apartridge, line 132: I think it is strange that the text here is quite different than what we have in C++. Fex the C++ text goes into more detail about calibration, while this text taks abound rounding etc. Is there a reason for it to be different? I think they should be identical except that C++ specific things must of course be replaced with Python.

test/test_projection.py
eskaur: Unfortunately these tests can only be run on real camera, so no file camera and no run in CI.

src/Projection.cpp
eskaur: Like this? Seems to work fine.
https://github.com/zivid/zivid-python/pull/234/commits/1a1eb23f0e102259c0a94f28eafb4f5c021028db

test/test_projection.py
eskaur: For now I think we just have to run the sample before each release and see that it looks fine.

modules/zivid/experimental/projection.py
eskaur: I could rename it to `show_image_bgra`. What do you think?

modules/zivid/experimental/projection.py
eskaur: I think it happened because I didn't like the C++ docstring very much.
Anyway, I agree consistency matters also. I'll change the main body to be identical:
```
This function takes 3D points in the camera's reference frame and converts them to the projector frame
using the internal calibration of a Zivid camera. In a Zivid point cloud, each point corresponds to a
pixel coordinate in the camera, but the projector has a slight offset. The translation of each point
depends on the distance between the camera and the point, as well as the distance and angle between the
camera and the projector.
```

Regarding this stuff:
```
This function takes 3D points in the camera's reference frame and converts them to the
projector's 2D (XY) frame using the internal calibration of a Zivid camera. The returned
points maps to projector pixels as round(X)->col and round(Y)->row.
```
it's something I added because it is very easy to get confused when we return (X,Y) but then index images as (row, col). Want me to remove it?

eskaur: Fixup that makes it consistent with C++:
https://github.com/zivid/zivid-python/pull/234/commits/e39d99e1732dabd9492395aba807f31f9f9bea0c

modules/zivid/experimental/projection.py
eskaur: Or leave it like this since it's Experimental.

modules/zivid/experimental/projection.py
apartridge: You can consider then to make a PR with your improvements to the C++ doc, to make it consistent? The main point is to make them consistent, but it can also be that C++ docstring can be improved.

eskaur: Let's merge this with consistency then (current fixup). Then we can always improve both for next release.

modules/zivid/experimental/projection.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __str__(self):
+        return str(self.__impl)
+
+    def capture(self, settings2d):
+        """Capture a single 2D frame without stopping the ongoing image projection.
+
+        This function can only be used with a zero-brightness 2D capture, otherwise it
+        will interfere with the projected image. An exception will be thrown if settings
+        contains brightness > 0 or if brightness is not set at all.
```
apartridge, line 43: This is also slightly out of sync with C++ version, fex C++ includes "This method returns right after the acquisition of the image is complete." Can you also check the other docstrings also, and either tweak them in the C++ or here so we get them the same?

modules/zivid/experimental/projection.py
```diff
@@ -0,0 +1,147 @@
...
+        camera._Camera__impl  # pylint: disable=protected-access
+    )
+
+
+def show_image(camera, image_bgra):
+    """Display a 2D color image using the projector.
+
+    This function returns a ProjectedImage instance. Projection will continue until this object is
+    destroyed or if its stop() method is called. Alternatively it can be used as a context manager,
+    in which case the projection will stop when exiting the "with" block.
```
apartridge, line 109: The c++ docstring mentions that this projection will also be cancelled if another capture is instantiated on the camera, but that is not mentioned here.

modules/zivid/experimental/projection.py
eskaur: Done, where it makes sense. The Python docstrings can't always be identical because "scope" is replaced with context manager stuff, etc.

modules/zivid/experimental/projection.py
eskaur: Did my best. Can't be identical, since Python is different from C++.

modules/zivid/experimental/projection.py
apartridge: I tried to find some other places in python sdk to follow as convention. We don't have anything exactly like this in the SDK (a variadic type as an input).

For point_cloud.copy_data it looks like ` point_cloud.copy_data(data_format="xyzrgba")`. Frame2D has two getters `frame_2d.image_rgba()` and `frame_2d.image_bgra()`. 

The most direct mapping to C++ SDK would be to have a this accept a  Image wrapper type. I do see that we have Image class, so the most C++ comparable fix would be to add a way to make an zivid-python Image, similar to what we added in C++ in 2.10. https://github.com/zivid/zivid-python/blob/36fd451396e3c92935bc8c9a84eb7be9a610c004/modules/zivid/image.py But we would get some complexity since as of now the C++ API will only accept a `Image<ColorBGRA>`. 

I think `show_image_bgra` and having it like you have it now is okey, then we can improve on this when we at some point add more Image types for showImage in the C++ API, without breaking too much existing zivid-python code.

modules/zivid/experimental/projection.py
apartridge: Yes, they wont be identical since the Python specific stuff must be included as you say, so we should just aim at making them as similar as they can be, but still correct for each language. And then when we add C# documentation it will be identical to C++ and Python except for using C# specific terms/namings.

Will you make a PR to sdk master after merging this, fixing the inconsistencies then? It is fine that it is out of sync until 2.11 but I think best to fix now else we will forget.

modules/zivid/experimental/projection.py
eskaur: Renamed here:
https://github.com/zivid/zivid-python/pull/234/commits/72854f750e707060925abc4af4819a5ca0995a17

apartridge (dismissed):
LGTM, I did not test this myself though but if sample is working well then it is fine to merge. If you want more testers then maybe CSU can try out this sample.

eskaur:
> LGTM, I did not test this myself though but if sample is working well then it is fine to merge. If you want more testers then maybe CSU can try out this sample.

Should be fine. I also have another test-sample (not in this PR) for projecting an image (Zivid logo). I'll make sure that one also runs fine before merging.

apartridge (approved):

PR #236, opened 2023-08-25
Add python support for presets API

johningve:
The presets API added in Zivid SDK 2.9 allows the user to obtain settings presets that are fine tuned for each camera model for a number of different use-cases.

See samples/sample_presets.py for an example of how to use it.

### DIFF
--- a/modules/zivid/presets.py
+++ b/modules/zivid/presets.py
@@ -1,0 +1,136 @@
+"""Contains the Preset and Category classes."""
+import _zivid
+
+from zivid.settings import _to_settings
+from zivid.camera_info import CameraInfo, _to_internal_camera_info
+
+
+class Preset:
+    """3D settings preset.
+
+    Presets are pre-defined settings that are tuned for different camera models to perform optimally
+    under different conditions and use cases.
+
+    This class cannot be initialized directly by the end-user. Use the presets method on the
+    Category class to obtain a Preset instance.
+    """
+
+    def __init__(self, impl):
+        """Initialize Preset wrapper.
+
+        This constructor is only used internally, and should not be called by the end-user.
+
+        Args:
+            impl:   Reference to internal/back-end instance.
+
+        Raises:
+            TypeError: If argument does not match the expected internal class.
+        """
+        if not isinstance(impl, _zivid.presets.Preset):
+            raise TypeError(
+                "Unsupported type for argument impl. Got {}, expected {}".format(
+                    type(impl), _zivid.presets.Preset
+                )
+            )
+        self.__impl = impl
+
+    @property
+    def name(self):
+        """Get the name of the preset.
+
+        Returns:
+            The name of the preset.
+        """
+        return self.__impl.name()
+
+    @property
+    def settings(self):
+        """Get the settings of the preset.
+
+        The settings might change between different releases of the SDK. New presets might be added
+        and old ones might be removed. If having the exact same settings is desired, it is
+        recommended to save them to a YML file and load as needed.
+
+        Returns:
+                The settings of the preset.
+        """
+        return _to_settings(self.__impl.settings())
+
+    def __str__(self):
+        return str(self.__impl)
+
+
+class Category:
+    """Preset category.
+
+    This class cannot be initialized directly by the end-user. Use the categories function to
+    obtain a list of categories available for a camera.
+    """
+
+    def __init__(self, impl):
+        """Initialize Category wrapper.
+
+        This constructor is only used internally, and should not be called by the end-user.
+
+        Args:
+            impl:   Reference to internal/back-end instance.
+
+        Raises:
+            TypeError: If argument does not match the expected internal class.
+        """
+        if not isinstance(impl, _zivid.presets.Category):
+            raise TypeError(
+                "Unsupported type for argument impl. Got {}, expected {}".format(
+                    type(impl), _zivid.presets.Category
+                )
+            )
+        self.__impl = impl
+
+    @property
+    def name(self):
+        """Get the name of the category.
+
+        Returns:
+            The name of the category.
+        """
+        return self.__impl.name()
+
+    @property
+    def presets(self):
+        """Get the presets available in the category.
+
+        The settings might change between different releases of the SDK. New presets might be added
+        and old ones might be removed. If having the exact same settings is desired, it is
+        recommended to save them to a YML file and load as needed.
+
+        Returns:
+            The presets available in the category.
+        """
+        return [Preset(p) for p in self.__impl.presets()]
+
+    def __str__(self):
+        return str(self.__impl)
+
+
+def categories(model):
+    """Get available preset categories for the specified camera model.
+
+    A preset category contains a collection of presets optimized for one scenario or use case.
+
+    The settings might change between different releases of the SDK. New presets might be added
+    and old ones might be removed. If having the exact same settings is desired, it is recommended
+    to save them to a YML file and load as needed.
+
+    Args:
+        model: The model for the camera whose preset categories should be returned. This value can
+               be obtained from CameraInfo.model.
+
+    Returns:
+        The available categories for the specified camera model.
+    """
+    return [
+        Category(c)
+        for c in _zivid.presets.categories(
+            _to_internal_camera_info(CameraInfo(model=model)).model
+        )
+    ]
--- a/samples/sample_presets.py
+++ b/samples/sample_presets.py
@@ -1,0 +1,39 @@
+"""Presets sample."""
+from zivid import Application
+from zivid.presets import categories
+
+
+def _main():
+    app = Application()
+
+    with app.connect_camera() as camera:
+        available_categories = categories(camera.info.model)
+        print("The following preset categories are available:")
+        for index, category in enumerate(available_categories):
+            print("    {}: {}".format(index, category.name))
+        chosen_category = available_categories[
+            int(input("Choose a category (enter a number): "))
+        ]
+
+        print(
+            "The following presets are available in category '{}':".format(
+                chosen_category.name
+            )
+        )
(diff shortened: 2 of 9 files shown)


modules/zivid/presets.py
```diff
@@ -0,0 +1,144 @@
...
+    if model not in CameraInfo.Model.valid_values():
+        raise ValueError(
+            "Invalid camera model. Got {}, expected one of {}".format(
+                model, CameraInfo.Model.valid_values()
+            )
+        )
+    # categories wants a _zivid.CameraInfo.Model. We need an _zivid.CameraInfo.Model.enum to
+    # construct it.
+    _model = _zivid.CameraInfo.Model(getattr(_zivid.CameraInfo.Model, model))
+    return [Category(c) for c in _zivid.presets.categories(_model)]
```
Ayxan13, line 144: I don't like using `getattr` here as it relies on names of the fields being the same as their values. How about this:
```suggestion
    return [
        Category(c) for c in _zivid.presets.categories(_to_internal_camera_info(CameraInfo(model=model)).model)
    ]
```

modules/zivid/presets.py
```diff
@@ -0,0 +1,144 @@
...
+        Args:
+            impl:   Reference to internal/back-end instance.
+
+        Raises:
+            TypeError: If argument does not match the expected internal class.
+        """
+        if not isinstance(impl, _zivid.presets.Category):
+            raise TypeError(
+                "Unsupported type for argument impl. Got {}, expected {}".format(
+                    type(impl), type(_zivid.presets.Category)
```
Ayxan13, line 84: `type(_zivid.presets.Category)` is similarly wrong here.

modules/zivid/presets.py
```diff
@@ -0,0 +1,144 @@
...
+        Args:
+            impl:   Reference to internal/back-end instance.
+
+        Raises:
+            TypeError: If argument does not match the expected internal class.
+        """
+        if not isinstance(impl, _zivid.presets.Preset):
+            raise TypeError(
+                "Unsupported type for argument impl. Got {}, expected {}".format(
+                    type(impl), type(_zivid.presets.Preset)
```
Ayxan13, line 32: `type(_zivid.presets.Preset)` is wrong here (it's always type). You want just `_zivid.presets.Preset`.

modules/zivid/presets.py
Ayxan13: Also, consider using fstrings instead of string.format

Ayxan13 (changes requested):

modules/zivid/presets.py
eskaur: I would just write "expected zivid.Preset" in the string actually. Something has to be hardcoded anyway, and `_zivid.presets.Preset` inside format() isn't any better.

samples/sample_presets.py
```diff
@@ -0,0 +1,35 @@
...
+    for index, preset in enumerate(chosen_category.presets):
+        print("    {}: {}".format(index, preset.name))
+    chosen_preset = chosen_category.presets[
+        int(input("Choose a preset (enter a number): "))
+    ]
+
+    print("Capturing point cloud with preset '{}' ...".format(chosen_preset.name))
+    with camera.capture(chosen_preset.settings) as frame:
+        frame.save("result.zdf")
+
```
eskaur, line 32: Love this sample. Maybe also add this at the end:

```
settings_file = chosen_preset.name + ".yml"
print(f"Saving Settings to {settings_file}")
chosen_preset.settings.save(settings_file)
```

modules/zivid/presets.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class Preset:
+    """3D settings preset.
+
+    Presets are pre-defined settings that are tuned for different camera models to perform optimally
+    under different conditions and use cases.
+
+    This class cannot be initialized directly by the end-user. Use the presets method on the
+    Category class to obtain a Camera instance.
```
eskaur, line 15: Do you mean "to obtain a Preset instance"?

modules/zivid/presets.py
eskaur: See the other similar comment for my opinion.

Regarding f-strings: We currently have no f-strings in this package at all. The reason is that we used to support Python 3.5, which doesn't have f-strings. Let's just keep it consistent for now, and then we can upgrade to f-strings later.

modules/zivid/presets.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __str__(self):
+        return str(self.__impl)
+
+
+def categories(model):
+    """Get available preset categories for the specified camera model.
+
+    A preset category contains a collection of presets optimized for one scenario or use case.
+    scenarios.
```
eskaur, line 119: Stray word "scenarios" after sentence.

test/test_presets.py
```diff
@@ -0,0 +1,15 @@
+def test_presets(application):  # pylint: disable=unused-argument
+    import zivid
+
+    categories = zivid.presets.categories(zivid.CameraInfo.Model.zividTwo)
```
eskaur, line 4: Maybe just iterate the test over all `zivid.CameraInfo.Model.valid_values()`.

modules/zivid/presets.py
johningve: This __init__ function is almost an exact copy of the init functions in `camera.py` and others. I guess they should all be changed. Maybe in a follow up PR?

Ayxan13: > Something has to be hardcoded anyway

@eskaur If the class gets renamed later no, your string will be quietly wrong but the format version will complain. I vote for using a runtime formatting, the overhead should be negligible.

Ayxan13: > I guess they should all be changed. Maybe in a follow up PR?
@johningve Yes, probably. But do fix this particular instance before merging, we shouldn't merge it with the bug :)

apartridge:
We also need to double-check that we are able to add 2D settings presets in the future when we want to without breaking any existing stuff. In the C++ API it is covered since CategoryBase and PresetBase are templated types, and I think it would also work quite well here since Python is very dynamic :)

johningve:
> We also need to double-check that we are able to add 2D settings presets in the future when we want to without breaking any existing stuff. In the C++ API it is covered since CategoryBase and PresetBase are templated types, and I think it would also work quite well here since Python is very dynamic :)

Yeah I don't think there will be any problem. It looks to me like adding 2D settings presets would add new types for 2D presets and categories, and would also require a new function to get the categories. So that could all be added without any conflict to the python wrapper. We will probably be able to reuse code for the preset and category classes.

modules/zivid/presets.py
johningve: Interestingly, type() does change the output of this error:

With type()

```
TypeError: Unsupported type for argument impl. Got <class 'list'>, expected <class 'pybind11_builtins.pybind11_type'>
```

Without type()
```
TypeError: Unsupported type for argument impl. Got <class 'list'>, expected <class '_zivid._zivid.presets.Category'>
```

I don't know why `_zivid` gets repeated here

johningve: > > Something has to be hardcoded anyway
> 
> @eskaur If the class gets renamed later on, your string will be quietly wrong but the format version will complain. I vote for using runtime formatting version, the overhead should be negligible.

:+1:

test/test_presets.py
```diff
@@ -0,0 +1,15 @@
...
+
+    category = categories[0]
+    assert isinstance(category, zivid.presets.Category)
+    assert len(category.name) > 0
+    assert len(category.presets) > 0
+
+    preset = category.presets[0]
+    assert isinstance(preset, zivid.presets.Preset)
+    assert len(preset.name) > 0
+    assert isinstance(preset.settings, zivid.Settings)
```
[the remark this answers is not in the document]
johningve: > `assert container` could be a nicer way of doing `assert len(container) > 0`

Ah I see that an empty list evaluates to false. Now that makes sense.

johningve:
So, there are two things to fix in a follow-up PR:

1. Remove the `type(_zivid....)` in the error thrown by initializers for certain classes
2. Scope the camera in the samples

johningve:
Ready for next round

samples/sample_presets.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    for index, preset in enumerate(chosen_category.presets):
+        print("    {}: {}".format(index, preset.name))
+    chosen_preset = chosen_category.presets[
+        int(input("Choose a preset (enter a number): "))
+    ]
+
+    print("Capturing point cloud with preset '{}' ...".format(chosen_preset.name))
+    with camera.capture(chosen_preset.settings) as frame:
+        frame.save("result.zdf")
```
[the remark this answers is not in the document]
eskaur: Then all the other samples are broken as well. I suggest we keep it consistent for now, and then SDK team can follow up with a PR that improves them all in one go.

But @apartridge can decide on this tradeoff with consistency.

modules/zivid/presets.py
eskaur: Okay I agree :)

eskaur (approved):
From my perspective this is fine given that the follow-up PRs also get done. Nice work :+1:

Ayxan13 (approved):
Looks good to me. I would scope the `Application` instance in the sample, but it's not a deal-breaker since we scopped the camera. It would be nice if one of the static linters could warn when objects with `__exit__` method are used without the `with` statement.

samples/sample_presets.py
apartridge: @FYI @SatjaSivcev, we are adding a presets sample for Python here, we may want to port this over to https://github.com/zivid/zivid-python-samples.
Can CSU make an issue for this?

PR #238, opened 2023-08-31
Add new in-field correction functions

johningve:
This commit exposes two new functions from the experimental in-field correction C++ API:

1. A new overload of detect_feature_points that takes a Frame as an argument. The existing python function has been updated to support both a Camera (the previously supported argument type) and a Frame.
2. capture_calibration_board, which captures a calibration board with the correct settings and returns the frame.

Fixes ZIVID-6764

### DIFF
--- a/modules/zivid/experimental/calibration.py
+++ b/modules/zivid/experimental/calibration.py
@@ -2,7 +2,9 @@
 
 import _zivid
 from zivid.calibration import DetectionResult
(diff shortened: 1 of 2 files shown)


johningve:
I haven't updated any samples / tests for this. Only limited manual testing to ensure that the changed functions work. I'm open for suggestions for sample code or how to test this.

PR #241, opened 2023-10-05
CI: disable GPG check for OneAPI repo on Fedora

johningve:
It seems like Intel have used the wrong key when building the their OneAPI packages. To fix CI we have to disable the GPG check until they have fixed their packages.

### DIFF
--- a/continuous-integration/linux/platform-dependent/common.sh
+++ b/continuous-integration/linux/platform-dependent/common.sh
@@ -18,12 +18,13 @@ function ubuntu_install_opencl_cpu_runtime {
 }
 
 function fedora_install_opencl_cpu_runtime {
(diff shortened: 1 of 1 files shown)


eskaur (approved):

eskaur:
Please make not to check if we can revert this in a weeks time or something.

apartridge:
Hmm, I think we should avoid removing the key check and rather wait a bit and see if the issue gets fixed. Also we should report the issue to them in case they are not aware. It is ok if zivid-python is red for a few days in my mind.

johningve:
Looks like they already fixed APT repositories, but probably haven't gotten around to fixing Yum repos yet. We can probably just wait until it fixes itself then.

PR #245, opened 2023-11-21
Add fallback instructions for building from source on Windows

eskaur:


README.md
```diff
@@ -50,6 +50,11 @@ If you are using a version of PIP older than version 19 please manually install
     cd zivid-python
     pip install .
 
+The above `pip install .` command may give permission issues on some Windows machines. If so, try the following instead:
+
+    python continuous-integration/windows/create_binary_distribution.py
+    pip install ./dist/*.whl
```
torbsorb, line 56: On my setup it keeps adding wheels to the dist folder. Wouldn't this command try to install all the wheels? Should probably instruct the user to clear the dist folder first?

vawale (approved):

README.md
eskaur: I suppose the assumption is that the user doesn't have a dist folder before running these two commands. In that case, there will be only one whl there.

vawale (approved):

PR #247, opened 2023-11-23
Fix unit-test fixture teardown

eskaur:
We noticed that unit-tests were failing when run with SDK 2.11. It turns out that we were not tearing down objects in the right order, and SDK 2.11 is less tolerance to such things.

This commit does two things to fix the unit-tests:
- Make sure that all fixtures that yield frames end up releasing those frames on teardown.
- Fix a test that created a new Application inside its code while also indirectly depending on the "application" fixture.

### DIFF
--- a/test/calibration/test_hand_eye.py
+++ b/test/calibration/test_hand_eye.py
@@ -95,7 +95,7 @@ def test_eyetohand_calibration_save_load(handeye_eth_frames, handeye_eth_poses):
         ]
     )
 
-    with tempfile.TemporaryDirectory() as tmpdir, zivid.Application() as _:
+    with tempfile.TemporaryDirectory() as tmpdir:
         transform = handeye_output.transform()
         file_path = Path(tmpdir) / "matrix.yml"
         zivid.Matrix4x4(transform).save(file_path)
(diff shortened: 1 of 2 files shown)


nedrebo:
> We noticed that unit-tests were failing when run with SDK 2.11. It turns out that we were not tearing down objects in the right order, and SDK 2.11 is less tolerance to such things.

Maybe mention this in the SKD changelog @apartridge? Just a reminder about the lifetime requirements for `Zivid::Applicaton`. Probably some customer code will hit the same and then it is nice if it is already addressed in the Changelog.

eskaur:
> > We noticed that unit-tests were failing when run with SDK 2.11. It turns out that we were not tearing down objects in the right order, and SDK 2.11 is less tolerance to such things.
> 
> Maybe mention this in the SKD changelog @apartridge? Just a reminder about the lifetime requirements for `Zivid::Applicaton`. Probably some customer code will hit the same and then it is nice if it is already addressed in the Changelog.

Honestly the docstring for Application is outrageously short

https://www.zivid.com/hubfs/softwarefiles/releases/2.6.0+fa0036c5-1/doc/cpp/classZivid_1_1Application.html#details

apartridge:
We want to first understand why there is a regression in 2.11 compared to 2.10. We will investigate this.

eskaur:
Actually the docstring in the header is good. But I can't find the same info on the doxygen page.

apartridge (approved):

PR #248, opened 2023-11-23
Add Ubuntu 23.10 to CI test matrix

eskaur:
We also make the linting job use this image, since we want that job to always use the latest supported Ubuntu.


nedrebo (approved):

nedrebo (commented):
>     # - The three latest Fedora releases that we can support

Unrelated, but curious what _"that we can support"_ means? Better to remove it?

eskaur:
> > ```
> > # - The three latest Fedora releases that we can support
> > ```
> 
> Unrelated, but curious what _"that we can support"_ means? Better to remove those words?

"Can support" means "we're able to get the image properly set up with OpenCL CPU runtime" :)

PR #251, opened 2023-12-12
Add PixelMapping

torbsorb:
Add wrapper for `zivid.calibration.pixel_mapping()` as well as `zivid.experimental.PixelMapping`

### DIFF
--- a/modules/zivid/experimental/_pixel_mapping.py
+++ b/modules/zivid/experimental/_pixel_mapping.py
@@ -1,0 +1,34 @@
+"""Module for experimental pixel mapping. This API may change in the future."""
+
+import _zivid
+
+
+class PixelMapping:
+    """Pixel mapping from subsampled to full resolution.
+
+    Required when mapping an index in a subsampled point cloud to e.g. a full resolution 2D image.
+    """
+
+    def __init__(self, row_stride=1, col_stride=1, row_offset=0.0, col_offset=0.0):
+        self.__impl = _zivid.PixelMapping(
+            row_stride, col_stride, row_offset, col_offset
+        )
+
+    @property
+    def row_stride(self):
+        return self.__impl.row_stride()
+
+    @property
+    def col_stride(self):
+        return self.__impl.col_stride()
+
+    @property
+    def row_offset(self):
+        return self.__impl.row_offset()
+
+    @property
+    def col_offset(self):
+        return self.__impl.col_offset()
+
+    def __str__(self):
+        return str(self.__impl)
--- a/modules/zivid/experimental/calibration.py
+++ b/modules/zivid/experimental/calibration.py
@@ -1,6 +1,7 @@
 """Module for experimental calibration features. This API may change in the future."""
 
 import _zivid
+from zivid.experimental import PixelMapping
 from zivid.calibration import DetectionResult
 from zivid.camera import Camera
 from zivid.camera_intrinsics import _to_camera_intrinsics
@@ -68,6 +69,35 @@ def estimate_intrinsics(frame):
         _zivid.calibration.estimate_intrinsics(
             frame._Frame__impl  # pylint: disable=protected-access
         )
+    )
+
+
+def pixel_mapping(camera, settings):
+    """Return pixel mapping information given camera and settings.
+
+    When mapping from a subsampled point cloud to a full resolution
+    2D image it is important to get the pixel mapping correct. This
+    mapping depends on camera model and settings. This function provides
+    the correct parameters to map the 2D coordinates in a point cloud
+    captured using `settings` to the full resolution of the camera.
+
+    Args:
+        camera: Reference to camera instance.
+        settings: Reference to settings instance.
+
+    Returns:
+        A PixelMapping instance.
+    """
+
+    pixel_mapping_handle = _zivid.calibration.pixel_mapping(
+        camera._Camera__impl,  # pylint: disable=protected-access
+        _to_internal_settings(settings),
+    )
+    return PixelMapping(
+        pixel_mapping_handle.row_stride(),
+        pixel_mapping_handle.col_stride(),
+        pixel_mapping_handle.row_offset(),
+        pixel_mapping_handle.col_offset(),
     )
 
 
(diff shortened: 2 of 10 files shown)


src/Calibration/Calibration.cpp
```diff
@@ -62,8 +62,16 @@ namespace ZividPython::Calibration
...
+                 [](ReleasableFrame &releasableFrame) {
+                     return Zivid::Experimental::Calibration::estimateIntrinsics(releasableFrame.impl());
+                 })
+            .def(
+                "pixel_mapping",
+                [](ReleasableCamera &releasableCamera, const Zivid::Settings &settings) {
+                    return Zivid::Experimental::Calibration::pixelMapping(releasableCamera.impl(), settings);
+                },
+                py::arg("camera"),
+                py::arg("settings"));
```
apartridge, line 75: This is experimental in C++, and thus should really be in a experimental namespace in Python. It seems we have done a bad mistake for the other ones abvove here by not separating Experimental to a separate namespace/module.

apartridge: Hmm I guess this is internal code only and that the public part is infact experimental. Then it looks Ok

modules/zivid/experimental/_pixel_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+            row_stride, col_stride, row_offset, col_offset
+        )
+
+    @property
+    def row_stride(self):
+        return self.__impl.row_stride()
+
+    @row_stride.setter
+    def row_stride(self, value):
+        self.__impl.row_stride = value
```
apartridge, line 23: Do we need these setters? We dont have that in C++? Just the consturctor.

test/experimental/test_pixel_mapping.py
```diff
@@ -0,0 +1,11 @@
...
+    from zivid.experimental.calibration import pixel_mapping
+    from zivid.settings import Settings
+
+    pixel_mapping_handle = pixel_mapping(
+        camera=file_camera, settings=Settings(acquisitions=[Settings.Acquisition()])
+    )
+    assert isinstance(pixel_mapping_handle.row_stride, int)
+    assert isinstance(pixel_mapping_handle.col_stride, int)
+    assert isinstance(pixel_mapping_handle.row_offset, float)
+    assert isinstance(pixel_mapping_handle.col_offset, float)
```
apartridge, line 11: We should also check the exact values that are used, IMO. To ensure that it is mapped correctly in the pybind wrapper.

src/Calibration/Calibration.cpp
eskaur: Yeah this is how we've done it before, with for example infield correction.

modules/zivid/experimental/_pixel_mapping.py
eskaur: I agree, leave it to only getters for now.

modules/zivid/experimental/_pixel_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+import _zivid
+
+
+class PixelMapping:
+    """Pixel mapping from subsampled to full resolution.
+
+    Required when mapping an index in a subsampled point cloud to e.g. a full resolution 2D image.
+    """
+
+    def __init__(self, row_stride=1, col_stride=1, row_offset=0, col_offset=0) -> None:
```
eskaur, line 12: We don't generally do type hints in this repo, so you can drop the "-> None".

modules/zivid/experimental/_pixel_mapping.py
torbsorb: https://github.com/zivid/zivid-python/pull/251/commits/65668230e040186f33ed889953a8bc14c70ac58e

modules/zivid/experimental/_pixel_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+        self.__impl = _zivid.PixelMapping(
+            row_stride, col_stride, row_offset, col_offset
+        )
```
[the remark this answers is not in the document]
torbsorb: https://github.com/zivid/zivid-python/pull/251/commits/bb420327c73bdffc3c4cfcb514adadacd5137054

test/experimental/test_pixel_mapping.py
torbsorb: https://github.com/zivid/zivid-python/pull/251/commits/bb420327c73bdffc3c4cfcb514adadacd5137054

modules/zivid/experimental/_pixel_mapping.py
torbsorb: https://github.com/zivid/zivid-python/pull/251/commits/1bc2593126b8ced67de2661e105a6b8e420020d7

modules/zivid/experimental/calibration.py
```diff
@@ -71,6 +72,35 @@ def estimate_intrinsics(frame):
...
+
+    Returns:
+        A PixelMapping instance.
+    """
+
+    pixel_mapping_handle = _zivid.calibration.pixel_mapping(
+        camera._Camera__impl,  # pylint: disable=protected-access
+        _to_internal_settings(settings),
+    )
+    return PixelMapping(
```
johningve, line 96: I think it is a bit strange that we here have a `_zivid.PixelMapping`, but we don't pass that into the constructor, instead the constructor creates another `_zivid.PixelMapping`.

torbsorb: It constructs and returns `zivid.experimental._pixel_mapping` which is mapped to `zivid.experimental.PixelMapping`.

torbsorb: There was a chicken and egg story here, right Eskil?

eskaur: The problem was that we really needed two constructors:

- One for the user that takes the four numbers
- One for us that takes a _zivid.PixelMapping

Since Python can't really overload constructors, we opted to only have the first one.
Open to better solutions, but I think this is fine.

johningve: Yeah, I'm fine with keeping this. We can find a cleaner solution in the future.

johningve (approved):
LGTM

eskaur (approved):

PR #252, opened 2023-12-12
Remove some code duplication 

vawale:
Removes code duplication in array2d and image class wrappers for Zivid::ColorRGBA and Zivid::ColorBGRA pixel formats. This helps when we add new pixel formats, for example, Zivid::ColorSRGB.

### DIFF
--- a/src/include/ZividPython/ReleasableArray2D.h
+++ b/src/include/ZividPython/ReleasableArray2D.h
@@ -1,8 +1,10 @@
 #pragma once
 
-#include <Zivid/PointCloud.h>
 #include <ZividPython/Releasable.h>
 #include <ZividPython/Wrappers.h>
+
+#include <Zivid/Color.h>
+#include <Zivid/PointCloud.h>
 
 namespace ZividPython
 {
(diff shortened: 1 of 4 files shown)


src/include/ZividPython/ReleasableArray2D.h
```diff
@@ -1,9 +1,11 @@
 #pragma once
 
-#include <Zivid/PointCloud.h>
 #include <ZividPython/Releasable.h>
 #include <ZividPython/Wrappers.h>
 
+#include <Zivid/Color.h>
+#include <Zivid/PointCloud.h>
```
eskaur, line 7: PointCloud.h already includes Color.h via Image.h, so I don't think this change is necessary.

vawale: Ah, clangd decided to add it for me. Its not necessary, but maybe its okay to have it explicit?

eskaur: Sure, no problem.

eskaur (approved):

PR #253, opened 2023-12-12
Support for Zivid SDK 2.11 and some new Image class functions

vawale:
Adds python wrappers for some new functions introduced in sdk release 2.11.


modules/zivid/point_cloud.py
```diff
@@ -62,6 +62,7 @@ def copy_data(self, data_format):
         z:          ndarray(Height,Width)   of float
         rgba:       ndarray(Height,Width,4) of uint8
         bgra:       ndarray(Height,Width,4) of uint8
+        srgb:       ndarray(Height,Width,4) of uint8
```
johningve, line 65: Can you add a test for this in test_point_cloud.py?

johningve (commented):
Looks good. I think there are some lint issues that will have to be resolved to get a green build. At least I think I recall fixing some things when making the latest preview branch.

modules/zivid/point_cloud.py
vawale: done

continuous-integration/linux/lint-cpp.sh
```diff
@@ -11,8 +11,8 @@ fi
 
 #Todo: buildDir=$1
 
-cppFiles=$(find "$ROOT_DIR" -name '*.cpp' |grep --invert-match src/3rd-party |grep --invert-match _skbuild)
-hFiles=$(find "$ROOT_DIR" -name '*.h' |grep --invert-match src/3rd-party |grep --invert-match _skbuild)
+cppFiles=$(find "$ROOT_DIR" -name '*.cpp' | grep --invert-match src/3rd-party | grep --invert-match _skbuild | grep --invert-match build)
```
johningve, line 14: I'm assuming this change was made so that you can have a local build folder to get intellisense in IDE?

vawale: yep :+1:

johningve (approved):
LGTM

eskaur (approved):

PR #255, opened 2023-12-12
Randomize Unit Test Execution Order

johningve:
This PR randomizes the order unit tests are executed in. In the process of enabling this I also fixed several order-dependency problems in the unit tests.


johningve:
Hmm, looks like pytest-randomly doesn't support Python 3.7 :(

johningve:
Python 3.7 is EOL since this summer. @eskaur should we drop support?

eskaur:
> Python 3.7 is EOL since this summer. @eskaur should we drop support?

@johningve 
Fine by me, but not really my call anymore. Two things:

- Check with @apartridge 
- Make sure that this repo no longer recommends Python 3.7 https://github.com/zivid/zivid-python-samples . (Maybe even remove that range and just refer to the readme in this repo for which python versions we support)

test/conftest.py
```diff
@@ -29,18 +29,24 @@ def file_camera_file_fixture():
...
+@pytest.fixture(name="file_camera", scope="function")
 def file_camera_fixture(application, file_camera_file):
     with application.create_file_camera(file_camera_file) as file_cam:
         yield file_cam
 
 
+@pytest.fixture(name="shared_file_camera", scope="module")
+def shared_file_camera_fixture(application, file_camera_file):
+    with application.create_file_camera(file_camera_file) as file_cam:
+        yield file_cam
```
eskaur, line 47: Was this done to speed things up?

johningve: Yes. My idea is that the `shared_file_camera` can be used for any test that doesn't need to mutate the camera (for example, by disconnecting it). But if a test needs to disconnect the camera, then it cannot used the shared fixture because it may break other tests that use the same fixture. The solution for that is to use a function scoped fixture, but that makes the tests run much slower because the function scoped file camera must be re-created for each test.

I think it increases the test duration by around 2-3x.

johningve:
@eskaur Two things are needed to bring back 3.7 and at the same time support 3.12:

1. `pytest-randomly` needs to be frozen at version 3.12. This version does not claim to support Python 3.12, but it appears to work.
2. `_version.py` needs to use a fallback on 3.7 since 3.7 doesn't have `importlib.metadata`.

johningve:
The previous PR changed `_version.py`

eskaur (dismissed):

eskaur (approved):

eskaur (approved):

PR #262, opened 2024-04-10
Fix flakiness in Windows CI workflow setup

johningve:
Windows CI workflows have been failing due to a failure to clean up temporary directories. This commit adds a workaround that ignores these errors.

When we in the future are able to use Python 3.10 features, we can pass `ignore_cleanup_errors=True` to `tempfile.TemporaryDirectory()` instead.

### DIFF
--- a/continuous-integration/windows/setup.py
+++ b/continuous-integration/windows/setup.py
@@ -6,6 +6,36 @@ from pathlib import Path
 from common import repo_root, run_process, install_pip_dependencies
 
 
+# On Windows the TemporaryDirectory cleanup can sometimes fail.
+# In Python 3.10 and newer, this class has a new argument called
+# `ignore_cleanup_errors` that would help us workaround this issue.
+# However, until we can use Python 3.10, we will need to implement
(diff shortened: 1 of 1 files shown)


continuous-integration/windows/setup.py
```diff
@@ -6,6 +6,36 @@
...
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            self._tmp_dir.cleanup()
+        except OSError:
+            pass
+        finally:
+            self._tmp_dir = None
+
+    def __fspath__(self):
+        return self.name
```
vawale, line 32: Is this method used anywhere?

vawale (approved):

continuous-integration/windows/setup.py
johningve: The AI suggested that I implement this so that it becomes possible to pass it to the constructor of `Path`. Looks like it is a method that is implemented by the `os.PathLike` interface. Didn't try without it though.

PR #274, opened 2024-06-25
Add `serialized` and `from_serialized` functions to datamodels

johningve:


modules/zivid/settings.py
```diff
@@ -2492,10 +2492,12 @@ def __str__(self):
     class Sampling:
         class Color:
             disabled = "disabled"
+            grayscale = "grayscale"
```
micragz, line 2495: It looks like the color settings ended up in the wrong commit?

johningve: Yeah, I'm gonna fix it when we start merging. Before merging all of these PRs I'll make a PR to switch over to the new SDK and update the data models

vawale (dismissed):

micragz (approved):

apartridge (approved):

PR #275, opened 2024-06-25
Add new non-experimental calibration board detection functions

johningve:
This commit adds wrappers for new calibration board detection functions in the non-experimental `calibration` module. The following functions are added:

`detect_calibration_board` - This function can be used with either a Frame or Camera as input, and returns a DetectionResult.

`capture_calibration_board` - This function can be used to capture a frame with a Camera that later can be passed as input to the former function.


johningve:
Merge this after #276 and use img01.zdf from that PR instead of adding a new ZDF here

micragz (approved):

PR #276, opened 2024-06-26
Add marker detection and hand-eye API wrappers

johningve:
This commit adds API wrappers for the SDK 2.13 fiducial marker detection and hand-eye calibration.

Added to the `zivid.calibration` module are the classes `MarkerShape`, `MarkerDictionary`, and `DetectionResultFiducialMarkers`, as well as the function `detect_markers`. The `DetectionResultFiducialMarkers` class can be used to construct a `HandEyeInput`, just like `DetectionResult`.

The hand-eye test data was updated with some newer ZDFs that were downsampled to save on Git LFS bandwidth.


vawale (approved):

torbsorb:
Please consider changing the format of `corners_in_pixel_coordinates` in order to simplify the following (which is currently required):

```python
rgba = camera_frame.point_cloud().copy_data("rgba")
rgb = rgba[:, :, :3].copy().astype(np.uint8)
if marker_list is not None:
    detected_markers = detection_result.detected_markers()
    detected_corners = [
        np.array(marker.corners_in_pixel_coordinates).reshape((4, 1, 2))
        for marker in detected_markers
    ]
    marker_ids = np.array([marker.id for marker in detected_markers])
    rgb = cv2.aruco.drawDetectedMarkers(rgb, detected_corners, marker_ids, [0, 255, 0])
```

johningve:
@eskaur @Frontier789 could you advise on this?

johningve:
Do we have the same issue in C++ then? Or is this just a problem in python?

eskaur:
> Do we have the same issue in C++ then? Or is this just a problem in python?

In C++ we return our own special types (std::array<Zivid::PointXY,4>), so the user has to do some conversion before passing it to OpenCV anyway. Therefore I don't consider this an issue in C++. In python it becomes more tempting to be super compatible with 3rd party APIs since we return universal stuff like lists and numpy arrays.

eskaur:
Working on making the corner getters return numpy arrays.

modules/zivid/_calibration/detector.py
```diff
@@ -68,6 +70,203 @@ def __str__(self):
...
+                valid_values().
+        """
+        if dictionary_name not in cls._valid_values:
+            raise ValueError(
+                "Invalid dictionary name '{}'. Valid values are {}".format(
+                    dictionary_name, cls.valid_values()
+                )
+            )
+
+        return cls._valid_values[dictionary_name].marker_count()
```
eskaur, line 213: Missing unit-tests for these

modules/zivid/_calibration/detector.py
```diff
@@ -68,6 +70,203 @@ def __str__(self):
...
+        """
+        return self.__impl.valid()
+
+    def allowed_marker_ids(self):
+        """Get the allowed marker ids this detection result was made with.
+
+        Returns:
+            A list of integers, equal to what was passed to the detection function.
+        """
+        return self.__impl.allowed_marker_ids()
```
eskaur, line 253: Missing test for this.

johningve: Added an assert in the existing test cases

eskaur (approved):

PR #277, opened 2024-06-27
Pass wrapper name when constructing Zivid::Application

RobKleinIkink:
- Modify Singleton to accept constructor arguments for the object it manages
- Use Zivid::Detail::createApplicationForWrapper to construct the Application for the Python wrapper

Part of ZIVID-8484

### DIFF
--- a/src/include/ZividPython/Releasable.h
+++ b/src/include/ZividPython/Releasable.h
@@ -1,9 +1,14 @@
 #pragma once
 
+#include <ZividPython/Traits.h>
+
+#include <pybind11/pybind11.h>
+
+#include <cstddef>
 #include <memory>
 #include <optional>
-#include <pybind11/pybind11.h>
 #include <stdexcept>
+#include <type_traits>
 
 #define WITH_GIL_UNLOCKED(...)                                                                                         \
     [&, this] {                                                                                                        \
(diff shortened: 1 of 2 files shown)


johningve (dismissed):
LGTM

johningve (approved):

src/include/ZividPython/Releasable.h
```diff
@@ -129,12 +129,13 @@ namespace ZividPython
     class Singleton
     {
     public:
-        Singleton()
+        template<typename... Args>
+        Singleton(Args &&...args)
```
vawale, line 133: we need to delete the copy/move ctors or restrict the type of `Args` to not be `Singleton`, else this constructor will take precedence over copy/move.

vawale: pushed a fixup for this

vawale (dismissed):

src/include/ZividPython/Releasable.h
```diff
@@ -129,12 +132,14 @@ namespace ZividPython
     class Singleton
     {
     public:
-        Singleton()
+        template<typename... Args,
+                 typename std::enable_if<!has_type<Singleton<T>, Args...>::value, void *>::type = nullptr>
```
vawale, line 136: Possibly also need to remove reference and cv qualifiers from Args

vawale: fixed

vawale (approved):

PR #280, opened 2024-07-05
Bump to Zivid SDK 2.13.1

johningve:
Fixed issue where the orientation of ArUco marker poses returned from `MarkerShape.pose` were inconsistent with calibration board poses. All poses from detected ArUco markers now point in the positive z-direction, i.e. away from the camera. ArUco marker poses were introduced in 2.13.0.


johningve:
@eskaur does this look right?

eskaur (approved):

PR #293, opened 2024-11-27
Zivid 8878 support nested data model types

johningve:
This PR updates the datamodel wrapper and frontend generator to support nested datamodels.

It won't build yet due to lacking `IsDataModelRoot` trait


src/include/ZividPython/DataModelWrapper.h
```diff
@@ -145,6 +145,40 @@ namespace ZividPython
...
+            class Helper
+            {
+                static constexpr auto concatenate() {
+                    constexpr auto prefix{ "_zivid." };
+                    constexpr size_t prefixLength{ std::char_traits<char>::length(prefix) };
+                    constexpr size_t nameLength{ std::char_traits<char>::length(T::name) };
+
+                    std::array<char, prefixLength + nameLength + 1> result{ '\0' };
+
+                    // TODO: Switch to std::copy_n when we can switch to C++20 and drop support for GCC 9.
```
vawale, line 161: nitpick: maybe mention that `std::copy` is not used because it is not `constexpr` before C++20?

continuous-integration/code-generation/datamodel_frontend_generator.py
```diff
@@ -520,9 +577,28 @@ def _parse_internal_datamodel(current_class: Any) -> NodeData:
             to_be_removed.append(child)
 
         elif child.is_leaf:
+            if child.underlying_zivid_class.value_type.startswith("_zivid"):
+                datamodel_name = child.underlying_zivid_class.value_type.split(".")[-1]
+                datamodel_name_snake = _to_snake_case(datamodel_name)
+                child.is_nested_datamodel = True
+                datamodel_snake_case = _to_snake_case(datamodel_name)
```
vawale, line 584: `datamodel_name_snake` and `datamodel_snake_case` are the same thing?

vawale (approved):
looks good to me. Would be easier to review with an example of nested datamodel type in python generated code :)

PR #294, opened 2024-11-27
Support wrapping uninstantiated datamodel nodes

johningve:
Previously we just hardcoded the Settings/Acquisition and Settings2D/Acquisition nodes. This commit makes it work for any uninstantiated node that is part of a datamodelList by recursively searching the datamodel.

### DIFF
--- a/src/include/ZividPython/DataModelWrapper.h
+++ b/src/include/ZividPython/DataModelWrapper.h
@@ -145,6 +145,9 @@ namespace ZividPython
             static constexpr const char *value{ TypeName<T>::value };
         };
 
+        template<typename DM, typename Dest, typename Node>
+        void findAndWrapUninstantiatedNodesInDataModel(Dest &dest, const Node &node);
+
         template<bool isRoot, typename Dest, typename Target>
         py::class_<Target> wrapDataModel(Dest &dest, const Target &target, const bool uninstantiatedNode = false)
         {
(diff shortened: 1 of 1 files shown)


src/include/ZividPython/DataModelWrapper.h
```diff
@@ -292,6 +290,64 @@ namespace ZividPython
...
+        static_assert(isDirectChild("", "TopLevel") == true);
+        static_assert(isDirectChild("", "TopLevel/WithChild") == false);
+        static_assert(isDirectChild("TopLevel", "TopLevel/WithChild") == true);
+        static_assert(isDirectChild("TopLevel", "TopLevel/WithChild/GrandChild") == false);
+
+        template<typename DM, typename PyDest, typename Node>
+        void findAndWrapUninstantiatedNodesInDataModel(PyDest &dest, const Node &node)
+        {
+            node.forEach([&](const auto &member) {
+                using MemberType = std::remove_const_t<std::remove_reference_t<decltype(member)>>;
```
vawale, line 323: maybe you also need to remove `volatile` qualifier, so `std::remove_cv_t` instead of `std::remove_const_t`.

src/include/ZividPython/DataModelWrapper.h
```diff
@@ -292,6 +290,64 @@ namespace ZividPython
...
+                    {
+                        // This node is instantiated.
+                        return;
+                    }
+
+                    // Check via path that the contained type is a direct child of the dest
+                    if constexpr(isDirectChild(DM::path, ValueTypeContained::path))
+                    {
+                        wrapDataModel<false>(dest, ValueTypeContained{}, true);
+                    }
```
vawale, line 347: this is okay for now. I wonder if it makes better sense to have `ValueTypeContained` defined as part of `DM::Descendants` for leafDataModelList types. For example, defining `using Settings::Acquisitions::Descendants = std::tuple<Settings::Acquisition>`

vawale (dismissed):
minor comments

src/include/ZividPython/DataModelWrapper.h
johningve: I guess I could also use `decay_t`

vawale: that works too, but I generally avoid it as it also performs array -> pointer conversion. That shouldn't be an issue in this case though.

johningve: Updated with `remove_cv_t`. I noticed that C++20 actually has a `remove_cvref_t` that we might be able to use in the future :)

vawale (approved):

PR #296, opened 2024-11-28
Add new capture methods capture2d/capture3d/capture2d3d

johningve:
These methods are intended to replace `capture` which is now deprecated.


modules/zivid/camera.py
```diff
@@ -47,9 +47,127 @@ def __str__(self):
     def __eq__(self, other):
         return self.__impl == other._Camera__impl
 
+    def capture2d3d(self, settings):
```
eskaur, line 50: Should we change the "samples" directory and README snippets to use these new functions?

johningve: Done

vawale (approved):

modules/zivid/camera.py
```diff
@@ -47,9 +47,128 @@ def __str__(self):
...
+            settings: Settings to use for the capture. Can be either a Settings2D instance or a Settings instance.
+
+        Returns:
+            A Frame2D containing a 2D image and metadata
+
+        Raises:
+            TypeError: If the settings argument is not a Settings2D or a Settings
+        """
+        if isinstance(settings, Settings2D):
+            return Frame2D(self.__impl.capture2D(_to_internal_settings2d(settings)))
```
apartridge, line 153: why using differen casing (2D) in the impl method?

johningve: Ah, I'll change it

modules/zivid/camera.py
```diff
@@ -48,9 +48,128 @@ def __str__(self):
...
+        """Capture a single 2D frame.
+
+        This method returns right after the acquisition of the images is complete, and the camera has stopped projecting
+        patterns. For more information, see the remarks section of `capture_2d_3d` above. Those remarks apply for both 2D,
+        3D, and 2D+3D captures.
+
+        This overload is provided for convenience. Note that only the Settings2D part under `Settings.color` will be
+        used for the capture. The other parts of the settings will be ignored.
+
+        This method will throw if `Settings.color` is not set.
```
apartridge, line 142: This method is used for capture_2d(settings) and capture_2d(settings_2d) it seems, this part of the docstring applies only for the former and not the latter. This could be clarified a bit in the docstring?

samples/sample_capture.py
```diff
@@ -14,7 +14,14 @@ def _main():
         settings.processing.filters.outlier.removal.enabled = True
         settings.processing.filters.outlier.removal.threshold = 5.0
 
-        with camera.capture(settings) as frame:
+        settings.color = Settings2D()
+        settings.color.acquisitions.append(Settings2D.Acquisition())
+        settings.color.acquisitions[0].aperture = 5.6
+        settings.color.acquisitions[0].exposure_time = datetime.timedelta(
+            microseconds=8333
+        )
```
apartridge, line 22: Did you check that we dont have same bug here as we had in .NET where we didnt mutate the actual settings.color but instead a copy?

test/test_camera_capture.py
```diff
@@ -1,6 +1,92 @@
...
+
+    acquisitions3d = [zivid.Settings.Acquisition()]
+    acquisitions2d = [zivid.Settings2D.Acquisition()]
+    settings = zivid.Settings(
+        acquisitions=acquisitions3d, color=zivid.Settings2D(acquisitions=acquisitions2d)
+    )
+
+    with shared_file_camera.capture_2d_3d(settings) as frame:
+        assert frame
+        assert isinstance(frame, zivid.frame.Frame)
```
apartridge, line 15: Could in this file test a bit more, such that the frame.settings matches what is set above, which would reduce chances of bugs in this code.

samples/sample_capture.py
johningve: Yeah. Added a commit to test this

modules/zivid/camera.py
johningve: Fixed

test/test_camera_capture.py
johningve: Fixed

apartridge (approved):

PR #298, opened 2024-12-03
Update checkerboard ZDFs and multicamera transforms

johningve:
In 2.14, we dropped support for legacy checkerboards that lack Aruco markers. This commit replaces the checkerboard frames which had legacy checkerboards with new frames. Also the multicamera transforms are updated based on the new frames.


vawale (approved):

johningve:
Will be included in #306

PR #303, opened 2024-12-05
Implement experimental point cloud export API

johningve:


modules/zivid/experimental/point_cloud_export/file_format.py
```diff
@@ -0,0 +1,185 @@
...
+            if value == PLY.Layout.unordered:
+                return _zivid.point_cloud_export.file_format.PLY.Layout.unordered
+            raise ValueError(
+                "Invalid layout '{}'. Valid layouts are: {}".format(
+                    value, cls.valid_values()
+                )
+            )
+
+    def __init__(
+        self, file_name, layout=Layout.ordered, color_space=ColorSpace.linear_rgb
```
vawale, line 80: default in sdk is updated to `ColorSpace.srgb` for all file formats.

vawale (approved):
looks good

vawale (approved):

modules/zivid/experimental/point_cloud_export/file_format.py
apartridge: It would be best if we could just inherit the default behavior from the SDK side?

johningve: We could do that. But then should we also mention the default in the comment? At least if we set the default in python also we can see that the documentation is correct.

vawale (approved):

PR #305, opened 2024-12-05
Limit ninja version due to bug in the latest version

johningve:
There is a bug in the latest ninja package in PyPI, version 1.11.1.2, that causes ninja to call itself recursively until the system runs out of memory. This commit limits ninja to the previous version, 1.11.1.1, to prevent the issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,6 +4,6 @@
 requires = [
     "cmake>=3.24.0",
     "conan>=2,<3",
-    "ninja",
+    "ninja<=1.11.1.1",
     "scikit-build",
     ]


vawale (approved):

dave992:
I am also running into this issue when running the `pip install` command. `ninja --version` keeps being called until the system runs out of memory. Unfortunately fixing the `ninja` version (or any of the other dependencies) did not seem to solve my issue.

We became aware of this problem 1-2 weeks ago.

Did you find any issue or merge request from `ninja` / `pip` / other relevant tool that is used where this behavior is reported? I only seem to find very old issues on the `ninja` issue tracker so far, although they do have some new comments:
https://github.com/ninja-build/ninja/issues/2187

PR #309, opened 2024-12-10
Improve error message when CMake is not found

johningve:


setup.py
```diff
@@ -81,7 +81,13 @@ def run_process(args, **kwargs):
     try:
         run_process(("cmake", "--version"))
     except Exception as ex:
-        raise RuntimeError(_make_message_box("!! CMake not found !!")) from ex
+        raise RuntimeError(
+            _make_message_box(
+                "!! CMake not found !!",
+                "Ensure PIP is updated to the latest version to handle build dependencies automatically,",
```
apartridge, line 87: LGTM, I would consider to prefix this with "Hint" and maybe some blank lines before and after just so this part is more visibile. Often people get overwhelmed by long error messages and won't see the tips/solution.

apartridge (approved):

setup.py
johningve: It will look like this: 
image

apartridge: looks good

PR #313, opened 2025-01-09
Unlock GIL in all releasable class methods

johningve:
In d64366eb, the wrapper macros in Releasable.h were updated to unlock the GIL when calling into the SDK. However, it seems a couple of macros were missed in this commit. This meant that functions like `Frame::point_cloud` and `Camera::state` could not run in parallel.

This commit adds the GIL unlocking to the wrapper macros that were missed and also add the GIL unlock to one function that does not use wrapper macros.

### DIFF
--- a/src/include/ZividPython/Releasable.h
+++ b/src/include/ZividPython/Releasable.h
@@ -25,7 +25,7 @@
 #define ZIVID_PYTHON_FORWARD_0_ARGS(functionName)                                                                      \
     decltype(auto) functionName()                                                                                      \
     {                                                                                                                  \
-        return impl().functionName();                                                                                  \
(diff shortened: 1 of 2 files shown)


johningve:
Note that some classes like the `Application` class or modules like `Projection` do not use releasable wrappers and so they don't do any GIL unlocking. I think we might want to find a different way to do GIL unlocking to ensure we don't end up missing it in some places.

vawale (approved):

vawale:
Consider adding a test that discovered this bug

johningve:
> Consider adding a test that discovered this bug

I don't think it will be easy to write such a test in python, due to requiring multithreading and the GIL.

